PR #79, opened 2015-04-19
Remove: SetXYOffset is not used.

Alberth289346:
Unused method, variables are always 0


LordAro:
It's not used now, but it could be used for other Viewport Widgets - hold off on this until we know for sure

PR #80, opened 2015-04-24
Rework fence storage

Alberth289346:
Fences did all kinds of dirty tricks to get sprites, positions, and voxels right. Cleaned it all up.

Only remaining problem is that the EdgeCursor cannot display a fence at a raised edge of a non-steep tile due to not having the slope below available. That will however become a complete non-issue by reworking the mouse modes.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -603,8 +603,7 @@ uint16 GetGroundFencesFromMap(const VoxelStack *stack, int base_z)
 }
 
 /**
- * Add/remove land border fence based on current land ownership for the given
- * tile rectangle.
+ * Add/remove land border fence based on current land ownership for the given tile rectangle.
  * @param x Base X coordinate of the rectangle.
  * @param y Base Y coordinate of the rectangle.
  * @param width Length in X direction of the rectangle.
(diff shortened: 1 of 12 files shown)


src/map.cpp
```diff
@@ -485,9 +495,115 @@ TileOwner VoxelWorld::GetTileOwner(uint16 x, uint16 y)
 	return this->GetStack(x, y)->owner;
 }
 
+/** Map of imploded base ground slope to mask of fence nibbles at the base voxel. */
+static const uint16 _fences_mask_at_base[] = {
+	0xFFFF, //  0
```
LordAro: what do these numbers represent?

src/map.cpp
```diff
@@ -485,9 +495,115 @@ TileOwner VoxelWorld::GetTileOwner(uint16 x, uint16 y)
...
+	uint8 slope = v->GetGroundSlope();
+
+	assert(slope < lengthof(_fences_mask_at_base)); // Top steep slopes are not allowed.
+	uint16 mask = _fences_mask_at_base[slope];
+	uint16 fences = v->GetFences() & mask;
+	if (HasTopVoxelFences(slope)) {
+		v = stack->Get(base_z + 1);
+		uint top_fences = (v != nullptr) ? v->GetFences() : ALL_INVALID_FENCES;
+		mask ^= 0xFFFF;
+		fences |= top_fences & mask;
```
LordAro: i think when magical bitshifting is done, i'd like some comments explaining what/why

src/map.h
```diff
@@ -114,35 +114,32 @@ struct Voxel {
...
-	 * sure to call this method for the right voxel in the voxel stack.
+	 * Set all the fences of the voxel.
+	 * @param fences Type of the fences at each edge.
 	 */
-	inline void SetFenceType(TileEdge edge, FenceType fence_type)
+	inline uint16 SetFences(uint16 fences)
 	{
-		assert(fence_type < FENCE_TYPE_COUNT || fence_type == FENCE_TYPE_INVALID);
-		SB(this->fence, (edge & 3) * 4, 4, fence_type & 0xF);
+		return this->fences = fences;
```
LordAro: Set\* functions don't normally return a value, and at a first glance this looks like a broken == comparison

src/tile.h
```diff
@@ -144,7 +146,7 @@ static inline bool IsImplodedSteepSlope(uint8 ts)
  */
 static inline bool IsImplodedSteepSlopeTop(uint8 ts)
 {
-	return ts >= 19;
+	return ts >= 15 + TS_TOP_OFFSET;
```
LordAro: 15 is still somewhat of a magic number..

Alberth289346:
fixes added, more docs about meaning of fence data, and what bit operations do, removed return value from SetFences, and added an enum for imploded slopes, in case one ever wonders what tiles exist.

Alberth289346:
savegame compati-dinges thingie added

Alberth289346:
one indent seems better for the mask array, thanks

ghost:
Just being particular. :p Everything else receives a :+1:.

src/map.cpp
```diff
@@ -411,6 +421,37 @@ void VoxelStack::Load(Loader &ldr)
...
+					if (this->voxels[i].GetGroundType() == GTP_INVALID) continue;
+					if (!IsImplodedSteepSlopeTop(this->voxels[i].GetGroundSlope())) continue;
+					uint16 mask = low_fences_mask[this->voxels[i].GetGroundSlope() - ISL_TOP_STEEP_NORTH];
+
+					/* Take out the fences of the top voxel that should be in the base voxel.
+					 * Make the low fences in the high voxel invalid. */
+					uint16 fences = this->voxels[i].GetFences();
+					uint16 lower_fences = fences & mask;
+					uint16 high_invalid = ALL_INVALID_FENCES & mask;
+					mask ^= 0xffff;
```
LordAro, line 445: I _could_ be exceptionally picky and say that it would be more consistent to use uppercase letters for the hex (as is usual for the rest of the program/prq) but I don't actually care that much :p 
However, I am vaguely concerned about the amount of code that's necessary here for loading "old fences". Are we that concerned about savegame compatibility at this point? If we are, these Load functions are quickly going to become very large, which scares me slightly

LordAro:
:+1:

PR #88, opened 2015-05-09
Implement game control actions

ghost:
In order to allow FreeRCT to correctly handle game initialization, loading, saving, and cleanup.
Note: Specific game data structure cleanup methods will be added in future commits.


ghost:
Suggested changes/updates made.

ghost:
I believe I've simplified the structure of my code with this latest rebase/commit. The 'game' code is now more focused on starting/loading/saving/closing a 'game level', which should help ease in the clean up of game objects (particularly VoxelObjects) in between 'levels/games'.

I am not saying that this code is final, but more so commenting to open up discussion in case I am still heading in the wrong direction with this PR. As always, I am open to any and all suggestions, critiques, etc. :)

ghost:
The GameControl class now implements 'action'-specific setup methods for preparing for the related GameControlAction (e.g., LoadGame is passed a file name, which it stores in the relevant member variable, and also sets the next action to be run to GCA_LOAD_GAME).

The main benefit of this class is it's RunAction method, which runs through the necessary logic for a specified action, providing the ability to correctly shutdown and/or start new 'games'/'levels'. This will (hopefully) be of great use for correctly cleaning up a game/level's data structures prior to starting the newly-requested game/level.

I believe this 'actions'-based route to be a lot more useful and sensible than the prior implementations I tried.

As always, I am open to any thoughts, concerns, ideas. Thanks.

src/freerct.cpp
```diff
@@ -21,6 +21,7 @@
 #include "fileio.h"
 #include "gamecontrol.h"
 
+GameControl _game_control;
```
Alberth289346: ///< Game controller.

src/freerct.cpp
```diff
@@ -126,13 +127,12 @@ int freerct_main(int argc, char **argv)
 
 	InitMouseModes();
 
-	StartNewGame();
+	/// \todo Allow for loading directly from a saved game.
+	_game_control.NewGame();
 
 	/* Loops until told not to. */
 	_video.MainLoop();
```
Alberth289346, line 135: Stuff is out of balance here, _game_control starts a new game outside _Video, but it doesn't close down here.

Code becomes easier to understand if you keep actions that come in pairs at the same level. Please consider either moving _game_control.NewGame() inside _video, or having the close action here after _video as well.

src/gamecontrol.cpp
```diff
@@ -15,13 +15,128 @@
...
+{
+}
+
+/**
+ * Run latest game control action.
+ * @pre next_action should not be equal to #GCA_NONE.
+ */
+void GameControl::RunAction()
+{
+	assert(this->next_action != GCA_NONE);
```
Alberth289346: This forces the call pattern if (_game_control.next_action != GCA_NONE) _game_control.RunAction();

It's nicer if you don't force these details on the caller. Instead, add a wrapper method (inline in the .h file for example), where you handle this. The caller can just call stupidly _game_control.DoIt();, and the c++ compiler will inline the test and skip the call.

src/gamecontrol.cpp
```diff
@@ -15,13 +15,128 @@
...
+{
+	this->fname = fname;
+	this->next_action = GCA_LOAD_GAME;
+}
+
+/**
+ * Prepare for a #GCA_SAVE_GAME action.
+ * @param fname Name of the file to write.
+ */
+void GameControl::SaveGame(const char *fname)
```
Alberth289346: Both for LoadGame and SaveGame, you are getting buffers filled with data here. The address gets copied and used, but releasing the buffer content isn't handled or documented.

If you want to keep the char *, I think it's fair if you make a copy during the call. That way, the caller can delete the buffer after the call returns. Another option is to not pass the pointer, but to take ownership of the buffer here, in which case this class must dispose of it at some point.
A third option is to use a std::string here, which does all the magic for you :)

src/gamecontrol.h
```diff
@@ -10,13 +10,102 @@
...
+	GameControl();
+	~GameControl();
+
+	void RunAction();
+	void NewGame();
+	void LoadGame(const char *fname);
+	void SaveGame(const char *fname);
+	void QuitGame();
+
+	bool running;
```
Alberth289346: ///< Indicates whether a game is running currently.

Alberth289346: Hmm, then the constructor is wrong, perhaps something else should be written here?

src/gamecontrol.h
```diff
@@ -10,13 +10,102 @@
...
+	~GameModeManager();
+
+	void SetGameMode(GameMode new_mode);
+	inline GameMode GetGameMode() const;
+	inline bool InPlayMode() const;
+	inline bool InEditorMode() const;
+
+private:
+	GameMode game_mode; ///< Current game mode of the program.
+};
```
Alberth289346, line 91: :smile: omg, this looks so unbelievably bloated, an entire class for one boolean :D

Not your current problem though, it's safe to ignore this comment :)

ghost, line 91: I was about to reply, "Umm, all I did was move it to this file." :stuck_out_tongue:
I can definitely see merging this into game control at some point, but will ignore for now.

src/video.cpp
```diff
@@ -454,9 +447,8 @@ void VideoSystem::MainLoop()
 {
 	static const uint32 FRAME_DELAY = 30; // Number of milliseconds between two frames.
 	bool missing_sprites_check = false;
-	_finish = false;
 
-	while (!_finish) {
+	while (_game_control.running) {
```
Alberth289346: I wonder if this would be better written as for (;;) {
and exiting only at line 464.

src/video.cpp
```diff
@@ -465,7 +457,11 @@ void VideoSystem::MainLoop()
 		for (;;) {
 			if (HandleEvent()) break;
 		}
-		if (_finish) break;
+		/* If necessary, change the played game. */
+		if (_game_control.next_action != GCA_NONE) {
+			_game_control.RunAction();
+		}
+		if (!_game_control.running) break;
```
Alberth289346, line 463: Could make it an inline method, not sure it's worth the effort for 1 boolean test.

Alberth289346:
Looking very good imho, nice progress

src/gamecontrol.cpp
ghost: Yikes, thanks for pointing this out, it appears I need to remember what language I'm using here...

src/gamecontrol.h
ghost: It may be worth having an Initialize method that sets this boolean, as well as handles starting a new game, saved game, etc. This would replace the _game_control.NewGame() call that is made prior to _video.MainLoop(). With the added Initialize(), I can also add an Uninitialize that would come after the MainLoop, which would most likely run ShutdownLevel for now.

Alberth289346: Sounds good to me

ghost:
I've implemented the latest suggestions. I will need to squash/fixup this commit into the previous one, but leaving them separate for now for easier comparison.

Thanks again for the comments and help.

~~Edit: Need to initialize running, next_action, and fname in constructor.~~

Alberth289346:
Looks good to me

ghost:
All suggestions/improvements have been added.

PR #89, opened 2015-05-10
Rework shop placer

Alberth289346:
Killing another old mouse mode, this time with a brand new shop placement window.


graphics/rcd/lang/en_GB.txt
```diff
@@ -170,6 +170,13 @@ strings {
 		COASTER_BUILD_BANK_RIGHT_TOOLTIP: "Selects track pieces that end with a banking to the right";
 		COASTER_BUILD_BUY_TOOLTIP:        "Click to buy the selected track piece";
 
+		// Ride build window strings.
+		RIDE_BUILD_TITLEBAR:              "Place the ride";
```
LordAro: Maybe "Place Ride" would be better for the title?

src/mouse_mode.h
```diff
@@ -214,7 +232,176 @@ class TileDataMouseMode : public MouseModeSelector {
...
+};
+
+/** %Tile data with ride voxel information. */
+struct RideTileData : public CursorTileData {
+	uint8 lowest;  ///< Lowest voxel in the stack that should be rendered.
+	uint8 highest; ///< Highest voxel in the stack that should be rendered.
+	std::vector<VoxelRideData> ride_info; ///< Ride information of voxels RideTileData::lowest to RideTileData::highest (inclusive).
+
+	/** Initialize #RideTileData and #CursorTileData data members. */
+	void Init() override
```
LordAro: doesn't quite match this patch, but why not constructors?

Alberth289346:
Replaced "Place the ride" titlebar text by "Place shop" (that is, specific for the kind of ride being placed), addressing the first (now deleted) comment.

The second comment about lack of constructors: The reason for not having constructors in the tile data, is because the data is initialized all the time. Every time the mouse is moved to another voxel, or when a different size or position of the cursor is forced by the window.

Adding a constructor would solve the first initialization only.

PR #91, opened 2015-05-30
Semi-transparent display of new shops while placing them.

Alberth289346:


Alberth289346:
Unduplicated the blend-pixel routine as first step

src/viewport.cpp
```diff
@@ -1199,6 +1218,7 @@ void PixelFinder::CollectVoxel(const Voxel *voxel, const XYZPoint16 &voxel_pos,
 			dd.base.x = this->rect.base.x - xnorth - x_off;
 			dd.base.y = this->rect.base.y - ynorth - y_off;
 			dd.recolour = nullptr;
+			dd.highlight = false;
```
LordAro, line 1221: perhaps it would be better to make dd.highlight = false the default? (set it in the constructor)

Alberth289346, line 1221: Yeah, except the struct doesn't have a constructor

PR #92, opened 2015-05-30
Display guest count in the bottom toolbar

Alberth289346:


src/person.cpp
```diff
@@ -1148,7 +1147,10 @@ bool Guest::DailyUpdate()
 
 	this->ChangeHappiness(happiness_change);
 
-	if (this->activity == GA_WANDER && this->happiness <= 10) this->activity = GA_GO_HOME; // Go home when bored.
+	if (this->activity == GA_WANDER && this->happiness <= 10) {
+		this->activity = GA_GO_HOME; // Go home when bored.
+		NotifyChange(WC_BOTTOM_TOOLBAR, ALL_WINDOWS_OF_TYPE, CHG_GUEST_COUNT, 0);
```
LordAro, line 1152: Why notify here? Guest number hasn't changed yet?

Alberth289346, line 1152: We don't have code detecting when he walks onto the first tile outside the park

I added a @todo somewhere to this fact

PR #93, opened 2015-06-01
DrawData constructor and simplification

Alberth289346:


src/viewport.cpp
```diff
@@ -224,6 +224,57 @@ class VoxelCollector {
  * @ingroup viewport_group
  */
 struct DrawData {
+	/**
+	 * Default constructor initializing only the fields that mostly always have the same value.
+	 */
```
LordAro: should be single line doxygen

src/viewport.cpp
```diff
@@ -224,6 +224,57 @@ class VoxelCollector {
...
+			this->level = dd.level;
+			this->z_height = dd.z_height;
+			this->order = dd.order;
+			this->sprite = dd.sprite;
+			this->base = dd.base;
+			this->recolour = dd.recolour;
+			this->highlight = dd.highlight;
+		}
+		return *this;
+	}
```
LordAro: can probably use `DrawData(...) = default;` for all 3 of these?

LordAro: Also, a bit out of scope of patch, but maybe z_height & base could be combined - we do have an XYZPoint type, after all
EDIT: Ah, probably not - pixels vs voxels

src/viewport.cpp
```diff
@@ -934,15 +945,9 @@ void SpriteCollector::CollectVoxel(const Voxel *voxel, const XYZPoint16 &voxel_p
 			FenceType fence_type = GetFenceType(fences, edge);
 			if (fence_type != FENCE_TYPE_INVALID) {
 				DrawData dd;
-				dd.level = slice;
-				dd.z_height = voxel_pos.z;
+				dd.Set(slice, voxel_pos.z, (edge + 4 * this->orient + 1) % 4 < EDGE_SW ? SO_FENCE_BACK : SO_FENCE_FRONT,
+						this->sprites->GetFenceSprite(fence_type, edge, gslope, this->orient), north_point);
```
LordAro, line 917: should be spaces instead of tabs? not sure, i always forget the coding style in these situations

ghost, line 917: Looks like the correct style is 2 additional tabs according to the "Methods and function implementation" section in the [CodeStyle](https://github.com/FreeRCT/FreeRCT/blob/wiki/CodeStyle.md):

"Simple statements like assignment or function calls that are very long can be split over multiple lines. In that case, add 2 additional TABs before the second and further lines to show the line is a continuation."

Alberth289346:
I don't really like the implicit initialization, so I added it to the DrawData::Set method instead.
Removed the various constructors, making the doxygen comment and the 'default' constructors obsolete as well.

PR #96, opened 2015-06-04
Add: Uninitialization of guests on level shutdown.

ghost:
This will also fix the game-crash behavior currently present when loading a saved game while guests are present (active) in the current game/level.

Note: I did not submit this as a 'Fix:' since it adds new things. :)

### DIFF
--- a/src/people.cpp
+++ b/src/people.cpp
@@ -79,6 +79,22 @@ Guests::~Guests()
 {
 }
 
+/** Deactivate all guests and reset variables. */
+void Guests::Uninitialize()
+{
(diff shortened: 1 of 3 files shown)


src/people.cpp
```diff
@@ -79,6 +79,22 @@ Guests::~Guests()
 {
 }
 
+/** Deactivate all guests and reset variables. */
+void Guests::Deinitialize()
```
LordAro: Uninitialize?

ghost:
Method renamed to Uninitialize.

PR #99, opened 2015-06-07
Doc: Doxygen fixes and additions.

Alberth289346:


Alberth289346:
Rewrote the '4 bit' check and comment. Also moved the check into a .cpp line to make Doxygen understand 'assert_compile'.

PR #104, opened 2015-06-14
Reimplement fence builder with mouse selector.

Alberth289346:
On top of PR #103.


src/map.h
```diff
@@ -666,19 +667,4 @@ static inline bool IsVoxelInsideWorld(const XYZPoint16 &vox)
...
 
-/**
- * Is both corners of given edge raised of a voxel with the
- * given non-steep slope?
- * @param edge The edge to check
- * @param slope The slope of a voxel.
- * @pre (slope & TSB_STEEP) == 0
- * @return True if both corners of given edge are raised, otherwise false.
- */
-static inline bool IsRaisedEdge(TileEdge edge, TileSlope slope)
```
LordAro, line 677: /me wonders if this function is useful elsewhere (or is already "used" elsewhere)

src/mouse_mode.h
```diff
@@ -404,6 +404,88 @@ class RideTileDataMouseMode : public TileDataMouseMode<TileData> {
...
+		VoxelTileData<VoxelRideData> &td = this->GetTileData(pos);
+		if (!td.cursor_enabled) return;
+		VoxelRideData &vrd = td.ride_info[pos.z - td.lowest];
+		vrd.sri = sri;
+		vrd.instance_data = instance_data;
+	}
+};
+
+/** Mouse mode displaying a cursor and fences. */
+class FencesMouseMode : public VoxelTileDataMouseMode<VoxelTileData<VoxelFenceData>> {
```
LordAro, line 450: Maybe "EdgeMouseMode" ?i can't think of anything at the moment, but it seems reasonable that there might be something other than fences that's placed on the edge of tiles

src/map.h
Alberth289346, line 677: It was used with the old fence builder code, as fences at a raised edge are in the voxel above the base level. That use got killed with new GetVoxelZOffsetForFence  function in commit 64a4f50 which generalizes the query. Currently I wouldn't know anything else that needs to know about raised edges. Paths need flat edges (ie both raised or both down), but that is more generic than this function. Most other code mostly cares about ground as "not be where I want to be" without caring about actual shape of the ground.

src/mouse_mode.h
Alberth289346, line 450: The idea is good, but it does "GetFences" rather than "GetEdges". At the calling side of that function, viewport pulls the original data from the "fences" value in the voxel. Sprites get pulled from "fences" part in the sprite collection. In other words, renaming does cascades through a whole lot of files, data structures, and functions. I think it should be done when we actually have other sprites than just fences.

PR #106, opened 2015-06-27
Doc: Doxygen fixes, and typo fix.

Alberth289346:
There are some more errors reported, but I think doxygen is wrong.


LordAro:
Which other errors are there?

LordAro:
goodenough.jpg

PR #107, opened 2015-07-07
Add: Cash spent by guest.

ghost:


LordAro:
I wonder if a "initial cash" attribute might be nicer (and const?) - then spent cash can just be worked out on the fly

ghost:
That could work. My reasoning was due to eventually having cash machines / ATMs, but it may be possible to have an initial_cash and cash (current_cash) and just increase both whenever a cash machine is visited.

LordAro:
oh, i hadn't thought of them. seems worse like that :)

ghost:
Yeah, I only took them into account since `ITP_MONEY` seems to indicate guests will eventually be able to obtain more money. :)

PR #108, opened 2015-07-25
Path build gui rewritten to new mouse selectors.

Alberth289346:
Deletion of old stuff isn't done yet, but a first attempt exploded somewhat, so it needs a little bit more care. Better do that in a new PR.


src/path_gui.cpp
```diff
@@ -320,62 +522,300 @@ void PathBuildGui::OnClick(WidgetNumber number, const Point16 &pos)
...
+		exits = GetDirections(this->build_pos);
+		directions = 0;
+		for (TileEdge edge = EDGE_BEGIN; edge < EDGE_COUNT; edge++) {
+			TileEdge reverse_edge = static_cast<TileEdge>((edge + 2) % 4);
+			XYZPoint16 neighbour(this->build_pos.x + _tile_dxy[edge].x, this->build_pos.y + _tile_dxy[edge].y, this->build_pos.z);
+			if (((0x01 << edge) & exits) != 0 && CanBuildPathFromEdge(neighbour, reverse_edge) != 0) {
+				directions |= 1 << edge;
+				continue;
+			}
+			if (((0x10 << edge) & exits) != 0) {
```
LordAro, line 596: arbitrary constant is arbitrary? maybe a name or at least a comment?

src/path_gui.cpp
```diff
@@ -320,62 +522,300 @@ void PathBuildGui::OnClick(WidgetNumber number, const Point16 &pos)
...
+		allowed_slopes = 0; // Slopes are disabled.
+	} else {
+		assert(this->build_pos.x >= 0); // Realized by 'disabled' above.
+		assert(this->build_direction != INVALID_EDGE);
+
+		TileEdge reverse_edge = static_cast<TileEdge>((this->build_direction + 2) % 4);
+		XYZPoint16 neighbour(this->build_pos.x + _tile_dxy[this->build_direction].x, this->build_pos.y + _tile_dxy[this->build_direction].y, this->build_pos.z);
+		if ((exits & (0x01 << this->build_direction)) != 0) {
+			allowed_slopes = CanBuildPathFromEdge(neighbour, reverse_edge);
+		} else if ((exits & (0x10 << this->build_direction)) != 0) {
```
LordAro, line 638: same here

PR #115, opened 2015-08-03
Change: Move to travis' container based infrastructure

LordAro:
Will probably need further changes before it works

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,18 @@
+sudo: false
 language: cpp
 
 compiler:
   - gcc
   - clang
 
-before_install:
-  # No SDL2 in the default repos yet
-  - sudo apt-add-repository ppa:zoogie/sdl2-snapshots -y
-  # GCC 4.8 not in the repos either
-  - sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
-  - sudo apt-get -qq update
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -38,11 +31,25 @@ env:
     secure: "gUgAXzGLyEkqSuMav5PLylFL3vtDO+TetIqzivkWDP5G6Md2W4xCSxBtgORtgOSngfJFipnoSvIzRo85ZyOJXKxN+1Ayy4D+xanFHpfY5pZtdhrpd9nbUXM/XLTJ9KeCd7n8BUq3Qd1h6cEAfCwvajSavC+bll3Df3MfSxISLdI="
 
 addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+      - debian-sid
+    packages:
```
ghost, line 38: I think we'll need `build-essential` here as well.

LordAro, line 38: no change, i don't think it's necessary

ghost:
Looks like it may be worth building the project in an Ubuntu VM using vagrant or something similar to test the SDL2-related issue. It seems to be installing the packages, but cmake isn't finding them for some reason. Anyways, just thoughts from my end. :)

Issue #56, opened 2015-03-16
Paths can be built through fences


GoogleCodeExporter:
```
It is possible to build a path through a fence or placing a fence on top of a 
path.

I have a work-in-progress patch for this, but it is not fully working yet.
```

Original issue reported on code.google.com by `leif.li...@gmail.com` on 17 Aug 2014 at 8:32

ghost:
@Leffe108: If it is all right, I've done some initial [work](https://github.com/FreeRCT/FreeRCT/compare/master...nicolasreed:path-fence-fixes) to fix this bug.

Feel free to let me know if this is an issue. :)

PR #116, opened 2015-08-04
Fix: Prevent - paths through fences and fences on paths.

ghost:
Probably needs a little bit added comment-wise before I'm finished...

Also, I'd like to wait for @Leffe108's response to https://github.com/FreeRCT/FreeRCT/issues/56#issuecomment-127734598 before merging this. :)


Alberth289346:
Just name it "paths and fences hate each other" :)

Looking good otherwise.

ghost:
Ha, that'd make for an interesting commit log :)

ghost:
Updated to include @Alberth289346's suggestion (quoted below):

> Untested, but all corners and edges go around in the same direction (except foundations somewhat), so a switch with 4 cases seems a bit overkill. Something like below perhaps?
> `if (GB(slope, PATHBIT_NE + edge, 1) != 0) edge = INVALID_EDGE;`
> 
> Other alternatives are pulling the value from an array like uint8 bitvalues[4]; or even uint32 bitvalues; (stuff the 4 bytes in a single 32bit integer).

PR #118, opened 2015-08-20
Save / load guests.

ghost:
This PR adds a number of methods and provides a few code changes in order to correctly save all guests currently in the world to the save game file, as well as correctly load the guests' data from the save game file.

I spent a fair amount of time with `git rebase` in order to (hopefully) have each commit make sense and keep specific additions/changes separate from one another; nevertheless, feel free to let me know if anything stands out as against this idea.

As you can see from the commit messages, a lot of credit has to go to @Alberth289346, especially in providing me with a lot of encoding/decoding code for `WalkInformation` and utf-8 strings.

As always, feedback is more than welcome.

Thanks.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -156,6 +156,35 @@ void VoxelObject::MarkDirty()
 	MarkVoxelDirty(this->vox_pos);
 }
 
+/**
+ * Load a voxel object from the save game.
+ * @param ldr Input stream to read.
+ */
+void VoxelObject::Load(Loader &ldr)
+{
+	uint32 x = ldr.GetLong();
+	uint32 y = ldr.GetLong();
+	uint32 z = ldr.GetLong();
+
+	XYZPoint32 xyz(x, y, z);
+
+	this->vox_pos = this->GetVoxelCoordinate(xyz);
+	this->pix_pos = this->GetInVoxelCoordinate(xyz);
+}
+
+/**
+ * Save a voxel object to the save game file.
+ * @param svr Output stream to write.
+ */
+void VoxelObject::Save(Saver &svr)
+{
+	XYZPoint32 xyz = this->MergeCoordinates();
+
+	svr.PutLong(xyz.x);
+	svr.PutLong(xyz.y);
+	svr.PutLong(xyz.z);
+}
+
 /** Default constructor. */
 VoxelStack::VoxelStack()
 {
--- a/doc/savegame.rst
+++ b/doc/savegame.rst
@@ -247,4 +247,85 @@ Version history
 
 - 1 (20150505) Initial version.
 
+
+Current guests block
+-------------
+The guests block stores all guests. Current version is 1.
+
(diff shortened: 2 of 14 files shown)


src/map.cpp
```diff
@@ -156,6 +156,35 @@ void VoxelObject::MarkDirty()
...
+	uint32 x = ldr.GetLong();
+	uint32 y = ldr.GetLong();
+	uint32 z = ldr.GetLong();
+
+	this->vox_pos.x = x >> 16;
+	this->vox_pos.y = y >> 16;
+	this->vox_pos.z = z >> 16;
+	this->pix_pos.x = x & 0xff;
+	this->pix_pos.y = y & 0xff;
+	this->pix_pos.z = z & 0xff;
```
LordAro: I wonder about a helper function for this - i believe it's done elsewhere
Same for the reverse operation below

Alberth289346: storing 8 bit values in 16 bits? nah :)
It would be in coaster cars I suspect, ie the other form of voxel objects.

Alberth289346:
yay! :)

ghost:
Thanks to the both of you for the suggestions; I'll let you know when 
I've implemented them.

ghost:
Implemented helper functions suggested by @LordAro.
Implemented other fixes/suggestions from @Alberth289346.

I'm not sure I named the helper functions in the best way, and the same can probably be said for the doxymentation for those functions. :)

I did not convert the one C-style cast to a `static_cast` since casting `uint*` to `char*` created more errors. We discussed utf8 strings a bit earlier today, but I am not sure whether I should delve into implementing more utf8 support for this PR, or leave that for another day/PR. :)

Thanks again for the help and thoughts.
As always, further comments/feedback are more than welcome.

ghost:
Updated.

Quoting @Alberth289346's one line note for the documentation change below since it may get lost in the rebase (not sure exactly how GitHub handles rebases and line notes in some instances).

> The doxymentation of the code should be independently readable from the code. In that way you get two independent descriptions, that you can check against each other.
> 
> The RCD format and the savegame format descriptions operate in the same way. They are an independent description of the meaning of the data, so you can cross-reference against the code.
> A second reason for having the RCD format and savegame format descriptions is that someone may at some point decide to write eg a second rcdgen program. Such a program must know what to deliver.
> 
> Here these worlds get mixed. Data is described in term of the implementation. You cannot write an implementation based on this document alone.
> I'll try to fix this.
> 
> Edit: How does https://paste.openttdcoop.org/ptfmmqc8l look ?

ghost:
Feel free to review this whenever you have the chance. :)

Alberth289346:
Was waiting for LordAro, as he couldn't decide whether he was done with the patch, but he has had his chance now :)

ghost:
Thanks! I was not sure whether my other comment had alerted the both of 
you that it was updated/finished, and wanted to make certain :)

PR #121, opened 2015-08-28
Refactor check data

Alberth289346:
Unify some for-loops into a single function, and make it simpler to see what blocks we actually have.


src/rcdgen/check_data.cpp
```diff
@@ -1117,6 +1117,35 @@ static std::shared_ptr<TCORBlock> ConvertTCORNode(std::shared_ptr<NodeGroup> ng)
 }
 
 /**
+ * @tparam Type to cast the element to.
```
ghost: Minor, but I think we'll want the parameter name, `T`, included for correct doxygen:
`@tparam T Type to cast the element to.`

ghost:
Looks good :+1:

Alberth289346:
fixed, thanks for noticing.

PR #123, opened 2015-10-08
Make freerct installable

vatriani:
implement install target for cmake


graphics/rcd/CMakeLists.txt
```diff
@@ -49,3 +49,7 @@ ENDFOREACH(RCDFILE)
 
 set_directory_properties(PROPERTIES ADDITIONAL_MAKE_CLEAN_FILES ${OUTFILES})
 add_custom_target(rcd ALL DEPENDS ${OUTFILES})
+
+# Install rules
+INSTALL(DIRECTORY DESTINATION share/FreeRCD)
```
LordAro: FreeRCD? --> FreeRCT ?

src/CMakeLists.txt
```diff
@@ -57,6 +60,7 @@ set(freerct_SRCS ${freerct_SRCS}
     "${CMAKE_SOURCE_DIR}/src/generated/coasters_strings.h"
     "${CMAKE_SOURCE_DIR}/src/generated/gui_strings.h"
     "${CMAKE_SOURCE_DIR}/src/generated/shops_strings.h"
+    "${CMAKE_SOURCE_DIR}/src/generated/comp_info.h"
```
LordAro, line 62: I have a feeling it might be worth merging rev and comp_info together - we hardly need 2 different tiny files for compile-time information

vatriani: rev implies reversion to me so it could  make  more sense to put them both together in compiling_information.h?

vatriani:
Squash it all together at 11be923cf9a1372440712026b26f34b4d31a27e5

ghost:
Looks good so far.

I'm wondering if we'll want to figure out a better place to put the freerct.cfg file since it would not be common practice to put such a file in a bin directory. Along with that, we will probably also want to come up with some kind of default freerct.cfg file to generate on install to make the process smoother for the end user.

Of course, there are a lot of important game elements that need implemented before worrying about such things; nevertheless, I feel as though this PR can't be merged until we figure out the .cfg stuff.

Thanks.

LordAro:
Right, yes, somehow managed to completely forget about that. IMO, the data files should have similar "levels" to how OTTD does it - http://hg.openttd.org/trunk.hg/file/7d0a8b844385/readme.txt#l267 (with the exception of not having ~/.freerct, i think) - and the cfg file should follow a similar structure (again, with ~/.config/freerct instead of ~/.freerct)

I'm afraid this is somewhat more work :)

src/CMakeLists.txt
LordAro, line 62: maybe compile_info.h (and .cpp? the nuances of how constant strings work still confuse me)
rev == revision, fyi ;)

PR #125, opened 2015-11-14
Add: Ability to load a saved game through the command line.

ghost:
This can hopefully be of some use to the project; it's been useful for my testing purposes so far.

I actually see it as a stepping stone towards being able to double click a saved game file in a folder and having the game launch with that saved game started.

I'm not sure whether or not it would be better to separate this into two commits since the Initialize method update could be seen as a separate change.

Thanks.


Alberth289346:
Two commits would have been a little cleaner, but the change is quite small.

PR #126, opened 2015-11-16
Add missing this to class members

ghost:
Only really necessary in order to adhere to the project's [coding style](https://github.com/FreeRCT/FreeRCT/blob/master/doc/CodeStyle.md#methods-and-function-implementation).

> Accesses to members of the class are always prefixed with `this->` to make them stand out with local variables.

Thanks.


Alberth289346:
Jolly good job! :)

PR #127, opened 2015-12-10
Saving and loading of rides

ghost:
I wouldn't call this _completely_ ready for merge just yet (although it does work fairly well), but I wanted to open it up for review to see if I'm headed in the right direction.

I'm mostly concerned with how I am detecting a RideInstance's RideType on Load. Currently, I have each instance save its RideType name so we can find and set the correct ride type on game load. This works in my general testing, but I could see it becoming an issue if RideTypes can have the same name, or a user ends up switching to an `.rcd` without one of the saved ride type names.

I could see saving the entire RideType object to the save game file as well, but it didn't feel right due to how we load RideTypes at the moment (using the RCD files).

Anyways, like I stated at the beginning, I still need to give this a review myself, but I wanted to gather others' input in case I'm missing anything critical (or going in a completely wrong direction).

Input is always welcome.

Thanks :smile:

### DIFF
--- a/src/coaster.cpp
+++ b/src/coaster.cpp
@@ -961,3 +961,88 @@ int CoasterInstance::GetNumberOfCars() const
 	if (number_cars == 0) number_cars = 1;
 	return number_cars;
 }
+
+void CoasterInstance::Load(Loader &ldr)
+{
+	this->RideInstance::Load(ldr);
+
+	this->capacity = (int)ldr.GetLong();
+	this->coaster_length = ldr.GetLong();
+
+	const CoasterType *ct = this->GetCoasterType();
+
+	uint saved_pieces = ldr.GetWord();
+	if (saved_pieces == 0) {
+		ldr.SetFailMessage("Invalid number of track pieces.");
+	}
+
+	for (uint i = 0; i < saved_pieces; i++) {
+		uint32 index = ldr.GetLong();
+		ConstTrackPiecePtr piece = ct->pieces[index];
+
+		if (piece != nullptr) {
+			this->pieces[i].piece = piece;
+			this->pieces[i].Load(ldr);
+			this->PlaceTrackPieceInWorld(this->pieces[i]);
+		} else {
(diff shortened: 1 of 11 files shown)


src/coaster.cpp
```diff
@@ -896,3 +937,68 @@ int CoasterInstance::GetNumberOfCars() const
...
+	uint saved_pieces = ldr.GetWord();
+	for (uint i = 0; i < saved_pieces; i++) {
+		uint32 index = ldr.GetLong();
+		this->pieces[i].piece = ct->pieces[index];
+		this->pieces[i].Load(ldr);
+		this->PlaceTrackPieceInWorld(this->pieces[i]);
+	}
+
+	int number_of_trains = (int)ldr.GetLong();
+	if (number_of_trains > 0) this->SetNumberOfTrains(number_of_trains);
```
LordAro: Maybe should fail in case of bad data, rather than just silently failing? Not sure if there's a facility for failing though...

Alberth289346: There is a Loader::SetFailMessage(), and a Loader::IsFail() or so, which seems to fit the bill.
It's denoted as for debugging and internal use, I guess technical failure reasons are not really something that a user can understand.  The "IsFail()" would still be usable I think, giving a general message like "Load failed as I don't get what the savegame says"  :p

Alberth289346:
Aside from the various details, looks like a good direction to me.

ghost:
Most (if not all) suggestions implemented.

I now save the `RideTypeKind` (shop, coaster, etc.) to use as an extra check in `RidesManager::Load` when looking for an existing valid `RideType` to use to create and load each `RideInstance` from the savegame file.

I'll also need to fail gracefully on bad data, as suggested by the both of you earlier; I am still looking at the best way to do so, though.

Any further suggestions, let me know.

Thanks!

ghost:
Updated. I added some extra checks for bad data, setting the Loader's fail message to something that reflects what went wrong during the load of said data.

Issue #128, opened 2015-12-16
CoasterTrains with more than one car do not correctly climb chain hills.


ghost:
I was testing out multiple trains and cars per train and noticed that multi-car trains do not 'climb' chained tracks correctly. I'm guessing this has to do with the way the `speed` is computed, possibly not taking into account that the second car is attached to the first or something along those lines.

Anyways, just wanted to make a note of this issue so that I wouldn't forget about it.

LordAro:
This doesn't surprise me in the slightest and I think your assessment is
correct, given the hill climb thing (and speeds in general) is just a hack
to make it work. There was no consideration for more than one car

I'm more impressed you've got more than one coastercar working! :)

ghost:
Yes, it seems to have to do with taking into account only one 
CoasterCar when
setting the speed of the CoasterTrain. I've made a temporary fix/hack 
on my end
that also checks to see if the current car has any cars in front of it 
that are
on a powered piece.

Once I clean up the fix, I'll make a pull request.

As for getting more than one CoasterCar working, I actually didn't have 
to do
anything special other than change the `max_number_trains` and
`max_number_cars` settings in the `tracks.txt` file, plus update the 
'default'
`Set` methods in `coaster.cpp`. I originally did this just to test my 
new save/load
code to see how it handled multiple trains/cars; of course, I then 
stumbled upon
this bug / not-yet-tested occurrence.

Yet another fun deviation from the original plan! :)

PR #129, opened 2015-12-17
Fix: Multi-car coaster trains correctly climb chain-lift hills. Fixes…

ghost:
… #128.


ghost:
~~Hmm, upon some further testing, this isn't fullproof yet.~~

ghost:
Ok, now it seems to be working. :)

PR #136, opened 2016-02-21
Add: Pan with GuiWindows open

CKraniak:
Changes:
- OnSelectorMoseMove() returns bool instead of void for all GuiWindows
  - Return value tells whether the mouse event should fall through to the viewport if the mouse is on the viewport
- Window manager gets new state: WMMM_MOUSE_DOWN
  - State causes all mouse move events to get routed to the "focus_holder", the window where the mouse was first clicked on, until the mouse button is released
- Point<T> now has operators
- _palette[] has a second way to specify the color without using magic numbers
  - This is used to remove the magic numbers in viewort.cpp
- Changes to PathGui to make it work more intuitively:
  - Right click still moves the viewport if it starts on a non-path
  - Right click goes into a "delete mode" and does not move the viewport if you start the right-click on a path
  - Fast mouse motion does not move the voxel highlight, nor does it cause a switch from "move" to "destroy" mode

### DIFF
--- a/src/coaster_gui.cpp
+++ b/src/coaster_gui.cpp
@@ -287,7 +287,7 @@ public:
 	void SetWidgetStringParameters(WidgetNumber wid_num) const override;
 	void OnClick(WidgetNumber widget, const Point16 &pos) override;
 
-	void SelectorMouseMoveEvent(Viewport *vp, const Point16 &pos) override;
+	bool SelectorMouseMoveEvent(Viewport *vp, const Point16 &pos) override;
 	void SelectorMouseButtonEvent(uint8 state) override;
 
 private:
@@ -626,23 +626,31 @@ void CoasterBuildWindow::SetupSelection()
 	this->piece_selector.pos_piece.piece = nullptr;
 }
 
-void CoasterBuildWindow::SelectorMouseMoveEvent(Viewport *vp, const Point16 &pos)
+/**
+ * Handles coaster build mouse movement events
+ * @param vp Viewport
+ * @param pos Mouse position
+ * @return True if the event no longer needs to be handled (currently always false)
+ */
+bool CoasterBuildWindow::SelectorMouseMoveEvent(Viewport *vp, const Point16 &pos)
 {
-	if (this->selector == nullptr || this->piece_selector.pos_piece.piece == nullptr) return; // No active selector.
-	if (this->sel_piece == nullptr || this->cur_piece != nullptr) return; // No piece, or fixed position.
+	if (this->selector == nullptr || this->piece_selector.pos_piece.piece == nullptr) return false; // No active selector.
+	if (this->sel_piece == nullptr || this->cur_piece != nullptr) return false; // No piece, or fixed position.
 
 	FinderData fdata(CS_GROUND, FW_TILE);
-	if (vp->ComputeCursorPosition(&fdata) != CS_GROUND) return;
+	if (vp->ComputeCursorPosition(&fdata) != CS_GROUND) return false;
 	XYZPoint16 &piece_base = this->piece_selector.pos_piece.base_voxel;
 	int dx = fdata.voxel_pos.x - piece_base.x;
 	int dy = fdata.voxel_pos.y - piece_base.y;
-	if (dx == 0 && dy == 0) return;
+	if (dx == 0 && dy == 0) return false;
 
 	this->piece_selector.MarkDirty();
 
 	this->piece_selector.SetPosition(this->piece_selector.area.base.x + dx, this->piece_selector.area.base.y + dy);
 	uint8 height = _world.GetTopGroundHeight(fdata.voxel_pos.x, fdata.voxel_pos.y);
 	this->piece_selector.SetTrackPiece(XYZPoint16(fdata.voxel_pos.x, fdata.voxel_pos.y, height), this->sel_piece);
+
+	return false;
 }
 
 void CoasterBuildWindow::SelectorMouseButtonEvent(uint8 state)
--- a/src/geometry.h
+++ b/src/geometry.h
@@ -37,7 +37,7 @@ struct Point {
 	 * @param y Y coordinate.
 	 */
 	template <typename OCT>
-	Point<CT>(Point<OCT>& that) :
+	Point<CT>(const Point<OCT>& that) :
 		x(static_cast<CT>(that.x)),
 		y(static_cast<CT>(that.y))
 	{
--- a/src/geometry.h
+++ b/src/geometry.h
@@ -29,10 +29,51 @@ struct Point {
 		this->y = y;
 	}
 
+	/**
+	 * Cast constructor. Allows conversion between 16-bit and 32-bit points.
+	 * @note This function does no bounds checking. The underlying types control the behavior of the cast; if you cast a 32-bit value to a 16-bit value, you get what you get.
+	 * @tparam OCT Other coordinate type
+	 * @param x X coordinate.
+	 * @param y Y coordinate.
+	 */
+	template <typename OCT>
+	Point<CT>(Point<OCT>& that) :
+		x(static_cast<CT>(that.x)),
+		y(static_cast<CT>(that.y))
+	{
+
+	}
+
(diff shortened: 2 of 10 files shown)


src/geometry.h
```diff
@@ -29,10 +29,51 @@ struct Point {
...
+	 * @note This function does no bounds checking. The underlying types control the behavior of the cast; if you cast a 32-bit value to a 16-bit value, you get what you get.
+	 * @tparam OCT Other coordinate type
+	 * @param x X coordinate.
+	 * @param y Y coordinate.
+	 */
+	template <typename OCT>
+	Point<CT>(Point<OCT>& that) :
+		x(static_cast<CT>(that.x)),
+		y(static_cast<CT>(that.y))
+	{
```
LordAro, line 43: Is this wanted? What's the usecase?

src/palette.h
```diff
@@ -13,6 +13,60 @@
...
+	PCSI_PINK_BROWN			= 10 + ( 8 * 12),
+	PCSI_DARK_PURPLE		= 10 + ( 9 * 12),
+	PCSI_BLUE				= 10 + (10 * 12),
+	PCSI_DARK_JADE_GREEN	= 10 + (11 * 12),
+	PCSI_PURPLE				= 10 + (12 * 12),
+	PCSI_RED				= 10 + (13 * 12),
+	PCSI_ORANGE				= 10 + (14 * 12),
+	PCSI_SEA_GREEN			= 10 + (15 * 12),
+	PCSI_PINK				= 10 + (16 * 12),
+	PCSI_BROWN				= 10 + (17 * 12)
```
LordAro: Broken alignment - needs spaces

src/palette.h
```diff
@@ -13,6 +13,60 @@
...
+};
+
+/**
+ * Get the index of the desired color in the initially-defined color palette.
+ * @param i Desired color
+ * @param n Desired intensity
+ * @return Index into _palette where the given color may be found
+ */
+inline int GetColorInFreeRCTPalette( PaletteColorStartIndex i,
+									 PaletteColorIntensity n) {
```
LordAro: newline?

src/path_gui.cpp
```diff
@@ -232,35 +248,93 @@ PathBuildGui::~PathBuildGui()
...
+		return IsInDeleteMode();
+	}
 
-	this->mouse_pos = fdata.voxel_pos;
-	if (this->single_tile_mode) this->TryAddRemovePath(_window_manager.GetMouseState());
+	// Move mode will sometimes let the voxel selection hop over to a
+	// neighboring voxel if the redraw isn't fast enough. If we don't skip the
+	// voxel position update, it will switch to a delete state in voxel hop
+	// casees, which is not desirable.
+	if (! IsInMoveMode()) this->mouse_pos = fdata.voxel_pos;
```
LordAro: extra space after '!'
several other places too

src/coaster_gui.cpp
```diff
@@ -626,23 +626,31 @@ void CoasterBuildWindow::SetupSelection()
...
 }
 
-void CoasterBuildWindow::SelectorMouseMoveEvent(Viewport *vp, const Point16 &pos)
+/**
+ * Handles coaster build mouse movement events
+ * @param vp Viewport
+ * @param pos Mouse position
+ * @return True if the event no longer needs to be handled (currently always false)
+ */
+bool CoasterBuildWindow::SelectorMouseMoveEvent(Viewport *vp, const Point16 &pos)
```
LordAro, line 635: Documentation for overridden functions is unnecessary - it's only being repeated from the declaration in window.cpp and doxygen manages it fine

src/window.cpp
```diff
@@ -919,18 +922,35 @@ void WindowManager::MouseMoveEvent(const Point16 &pos)
...
 	this->mouse_pos = pos;
 
+	// TODO: add a MOUSE_BUTTON_DOWN mode, send events only to the window that
+	// was corrent at the time the mode began.
 	switch (this->mouse_mode) {
 		case WMMM_PASS_THROUGH: {
+			/* Compute position relative to window origin. */
+			Point16 pos2;
+			if(this->current_window != nullptr)
+			{
```
LordAro: codestyle

src/viewport.cpp
```diff
@@ -671,7 +671,7 @@ void SpriteCollector::CollectVoxel(const Voxel *voxel, const XYZPoint16 &voxel_p
 			int x_off = ComputeX(vo->pix_pos.x, vo->pix_pos.y);
 			int y_off = ComputeY(vo->pix_pos.x, vo->pix_pos.y, vo->pix_pos.z);
 			Point32 pos(north_point.x + this->north_offsets[this->orient].x + x_off,
-			            north_point.y + this->north_offsets[this->orient].y + y_off);
+						north_point.y + this->north_offsets[this->orient].y + y_off);
```
LordAro: tabs

src/window.h
```diff
@@ -419,6 +419,24 @@ class WindowManager {
 		return this->viewport;
 	}
 
+	/**
+	 * Retrieve windows forcibly holding focus
+	 * @return The window forcibly holding the focus, or nullptr if none is forcibly holding focus
+	 */
+	inline Window * GetFocusHolder() const
```
LordAro: extra space

src/window.h
```diff
@@ -433,8 +451,9 @@ class WindowManager {
 
 	Point16 mouse_pos;        ///< Last reported mouse position.
 	Window *current_window;   ///< 'Current' window under the mouse.
+	Window *focus_holder;     ///< Window that's currently demanding events be routed to it
 	GuiWindow *select_window; ///< Cache containing the highest window with active GuiWindow::selector field
-	                          ///< (\c nullptr if no such window exists). Only valid if #select_valid holds.
+							  ///< (\c nullptr if no such window exists). Only valid if #select_valid holds.
```
LordAro: tabs

src/window.cpp
```diff
@@ -946,6 +966,26 @@ void WindowManager::MouseMoveEvent(const Point16 &pos)
...
+		case WMMM_MOUSE_DOWN: {
+			if((this->mouse_state & MB_RIGHT) != MB_RIGHT) {
+				this->mouse_mode = WMMM_PASS_THROUGH;
+				this->focus_holder = nullptr;
+				return;
+			}
+
+			Point16 pos2;
+			if(this->current_window != nullptr)
+			{
```
LordAro: codestyle

src/path_gui.cpp
```diff
@@ -232,35 +248,93 @@ PathBuildGui::~PathBuildGui()
...
+	// Check to see which RMB state we should be in.
+	// The whole purpose of checking our RMB state is to tell the Viewport
+	// whether it should process the event; we want the screen to hold still if
+	// we're deleting paths.
+	uint8 mouse_state = _window_manager.GetMouseState();
+	if((mouse_state & MB_RIGHT) == 0) {
+		this->mode = PS_RMB_MODE_NONE;
+	}
+	else {
+		if(! IsInDeleteMode()) {
```
LordAro: combine else with if ?
also, codestyle ;)

src/geometry.h
CKraniak, line 43: Use was in window manager's onmousemoveevent, which also changed to use this.

src/palette.h
CKraniak: I have a habit of keeping my lines shorter, and newlining to fix it if it gets too long.

src/coaster_gui.cpp
CKraniak, line 635: For the Path Gui override, return behavior is actually different. The others you may be correct about.

src/window.cpp
CKraniak: The open braces, right?

CKraniak:
Updated: rebased and fixed ... I think all of the formatting / style issues brought up.

CKraniak:
Note for anyone reading this comment chain:

A bunch of discussion occurred on IRC, and it was felt the responsibility for moving the viewport needed to be moved to the window manager and out of the viewport itself. I don't expect this to be merged until that is addressed.

PR #138, opened 2016-04-11
Return correct ride visit desire

ghost:
I'm not sure how useful this is at the moment, but I believe the intended behavior is to return the different `RideVisitDesire` values (`RVD_MAY_VISIT` versus `RVD_MUST_VISIT`).


ghost:
Discussed on IRC. Merging...

PR #141, opened 2016-04-18
Add: New GUI Graphics

KevDoy:
Restoring from backup


ghost:
Thanks for rebasing!

That's unfortunate that you had to recreate them; usually git has 
'backups',
or history as it were, somewhere in the .git directory.

Hopefully that won't happen again, but if so, there may be other ways 
to recover
from it.

Thanks again for your contributions.

KevDoy:
Yeah, no big deal. I appreciate the suggestions - checked the .git folder with no luck. Oh well 😄

PR #142, opened 2016-04-20
Port image processing tools to python to allow 32bpp sprite sheet rendering

paulpage:
I noticed that imagej does not support alpha channels, thus making it impossible to use the existing sprite sheet creation and masking scripts to render 32bpp sprite sheets. So I ported the macros to python. This allows rendering 32bpp sprite sheets with transparent backgrounds. 

Issues:
- 32bpp rendering of the tileselection sprites makes it hard to select corners when terraforming. 

Requirements to run scripts:
- Python3
- Pillow
- Tkinter

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -38,6 +38,7 @@ jjhdnd           GUI
 PovAddictW       FileIO
 stanekj          Rides
 tnt.freesoftware GUI
+Paul Page        Graphics
 ================ ========
 
 
--- a/graphics/sprites/info_and_scripts/mergeRenders.py
+++ b/graphics/sprites/info_and_scripts/mergeRenders.py
@@ -1,0 +1,62 @@
+"""
+This file is part of FreeRCT.
+FreeRCT is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+FreeRCT is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with FreeRCT. If not, see <http://www.gnu.org/licenses/>.
+"""
+
+from PIL import Image
+import os
+import math
+import argparse
+
+#tile widths to expect for renders
+tilew = (64, 128, 256)
+
+#length of string for tile index number
+indexleng = 5
+
+#parser for command line arguements
+parser = argparse.ArgumentParser(description = "Merge sprites in a directory into a sprite sheet.")
+parser.add_argument("fpath", help = "directory to work in")
+parser.add_argument("w", help = "columns in sprite sheet (default: 4)", type = int, nargs="?", default = "4")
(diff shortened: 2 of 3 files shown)


ghost:
@LordAro, if you wouldn't mind looking at this when you have a chance, 
I think this is more your territory than mine. :P

ghost:
@PJPage, thanks for the contribution.

Would you like to add a command line interface to these tools (in this PR), so that they can be used for scripting? If so, you could use the [argparse](https://docs.python.org/3.5/library/argparse.html) module.

As for the selecting of corners issue, that is due to the fact that the code currently checks for an 8bpp index.

Thanks.

paulpage:
Here's the command line interface. I left the old interface in there, but it can probably be deleted. 

Could you give me more detail about the code checking for an 8bpp index? Is this issue something that could be fixed simply by telling it to check for a 32bpp index, or is it more complicated than that? If it is that simple, where specifically is the code that checks for an 8bpp index? I checked in graphics/rcd/freerct.txt, mouse_mode.h/cpp, viewport.h/cpp terraform_gui.cpp and terraform.h/cpp, but I can't find it.

Alberth289346:
I didn't carefully studied your code, but if the code is not reachable, then yes, by all means, delete dead code. It probably also kills the Tkinter dependency, which is useful too.
Dead code tends to linger forever in a project. Nobody knows its purpose, since it's dead, and nobody dares to delete it, since it has unknown purpose!

Other things I noticed:
- There is a commit style that we use for commit messages. Yours would be something like "Add: Python image processing tools for 32bpp sprite sheets".
- New files must have the GPL license included in the header. Copy one from a cpp file, and change the comment prefix.
- You can add yourself as contributor in the contributor file if you like

The index check does in fact do a RGBA check, in viewport.cpp 1014 to 1033 doe detecting edges and corners (or the other way around), by

```
if (collector.pixel == _palette[181])
```

it tests for 4 very specific rgb colour triplets.

paulpage:
Ok, I made the required changes to the scripts.

I'll have to experiment with that viewport.cpp code to see if I can coax it into working with the 32bpp sprites.

ghost:
Thanks for making those changes.

One more thing, would you mind rebasing and squashing the second and 
third commits into the first one, and change the first commit's message 
to:
"Add: Python image processing tools for 32bpp sprite sheets."

Thanks!

paulpage:
I think that worked

ghost:
Hmm, the commit containing all the changes is there, but the other commits are still present, as well as a new 'Merge' commit, which shouldn't happen when doing a rebase in this case.

paulpage:
Sorry, I'm really inexperienced when it comes to git (If you couldn't tell :P). It looks like everything is squashed nicely now though.

ghost:
Hey, no worries, it seems you successfully rebased, so nice work. :)

I noticed the tkinter module is still in the spriteSheetMasking.py file.
Would you mind removing that line?

Also, could you add one more line to both python files at the end of 
the license section?
`See the GNU General Public License for more details. You should have 
received a copy of the GNU General Public License along with FreeRCT. 
If not, see <http://www.gnu.org/licenses/>.`
It's in all of the other files, so this keeps things consistent.

Other than that, looks good.

Thanks!

ghost:
All looks good now. Thanks again.

PR #144, opened 2016-10-09
Enable OSX testing on Travis CI.

ghost:
Travis CI will now build the project on both Linux and OSX.

### DIFF
--- a/.travis/before_install.sh
+++ b/.travis/before_install.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
+  brew update
+  brew install sdl2 sdl2_ttf
+else
+  # No SDL2 in the default repos yet
+  sudo apt-add-repository ppa:zoogie/sdl2-snapshots -y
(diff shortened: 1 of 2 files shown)


.travis/before_install.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
Alberth289346, line 1: The generally recommended practice is to do

``` sh
#!/usr/bin/env bash
```

This looks for 'bash' in the path, rather than assuming it's /bin/bash (which only holds for some Linuces, and usually not for all non-Linuces).

An alternative can be

``` sh
#!/bin/sh
```

Which is the default shell everywhere (although not sure it's actually required by some standard to exist).
While Linuces typically use 'bash' as 'sh' replacement, other systems (eg BSD) use the real sh shell. Your script thus may not contain bash-specific code in this case.

Since this code is specific for Travis, I see no real harm in doing it like you propose.

ghost:
Thanks for pointing that out; I have used the recommended practice
elsewhere, but completely forgot to use it in this script.

Anyways, thanks for the quick review and merge.

PR #148, opened 2019-06-19
Edit ride name

sgtbigman:
Add ability to edit a ride's name.

- Add `TextBuffer` class to support text editing.
- Add `EditTextWindow`:
  - Allows the player to manipulate the given text.
- Add 'Rename' button to ride window.

*Note*: **I primarily wrote this a couple of years ago when FreeRCT development was more active, so I'll need to clean up / update some things before this can be merged.**

Thanks!


src/coaster_gui.cpp
```diff
@@ -161,6 +164,11 @@ void CoasterInstanceWindow::OnClick(WidgetNumber widget, const Point16 &pos)
 		case CIW_OPEN:
 			this->SetCoasterStateForRadioButton(widget);
 			break;
+
+		case CIW_EDIT_NAME:
+			if (HighlightWindowByType(WC_EDIT_TEXT, ALL_WINDOWS_OF_TYPE)) return;
+			ShowEditTextGui(&this->ci->name, MAX_RIDE_INSTANCE_NAME_LENGTH);
```
Alberth289346, line 170: This looks horribly scary, a first search in the code suggests the text isn't actually copied anywhere, which means it's editing the memory block owned by the coaster. This is also somewhat confirmed by lack of a parameter so you can find the coaster instance again when editing is finished.
If this is indeed true, the problem with this solution is that you can create all kinds of nice crashes by opening a edit session and then delete the coaster instance, so you're editing unowned memory.

It has been too long ago to know if the above holds.

EDIT: Another point that needs to be checked, can you start 2 edits of the same coaster name?

src/text_buffer.h
```diff
@@ -0,0 +1,40 @@
...
+		const std::string &GetText() const;
+		void AppendText(const char *txt);
+		void InsertText(const char *txt);
+		void RemoveLastCharacter();
+		void RemovePrevCharacter();
+		void RemoveCurrentCharacter();
+
+		void SetPosition(int position) { this->current_position = position; };
+		const uint GetPosition() const { return this->current_position; };
+		void IncPosition() { if (this->current_position < this->text.length()) this->current_position++; };
```
Alberth289346:
```suggestion
		void IncPosition() { this->current_position = min(this->text.length() - 1, this->current_position + 1); };
```
and similarly for `DecPosition`

src/text_gui.cpp
```diff
@@ -0,0 +1,222 @@
...
+			int width, height = 0;
+			int start_index = 0;
+			int end_index = sb->GetVisibleCount();
+			if (this->text_buffer.GetPosition() < sb->GetVisibleCount()) {
+				start_index = 0;
+				end_index = this->text_buffer.GetPosition();
+			}
+			_video.GetTextSize((const uint8 *)this->text_buffer.GetText().substr(start_index, end_index).c_str(), &width, &height);
+			_video.BlitText((const uint8 *)"_", MakeRGBA(255, 255, 255, OPAQUE), GetWidgetScreenX(wid) + this->cursor_width + width, GetWidgetScreenY(wid) + (height / 2), this->cursor_width);
+			}
```
Alberth289346: This not dedenting can't be right, align it with the `c` of `case`.

src/text_gui.cpp
```diff
@@ -0,0 +1,222 @@
...
+	pt.y = (_video.GetYSize() - this->rect.height) / 2;
+	return pt;
+}
+
+void EditTextWindow::SetWidgetStringParameters(WidgetNumber wid_num) const
+{
+	switch (wid_num) {
+		case ETW_EDIT_TEXT: {
+			_str_params.SetUint8(1, (uint8 *)this->text_buffer.GetText().c_str());
+			}
```
Alberth289346: Doesn't need curly brackets

src/text_buffer.h
```diff
@@ -0,0 +1,40 @@
...
+
+		void SetText(const char *txt);
+		const std::string &GetText() const;
+		void AppendText(const char *txt);
+		void InsertText(const char *txt);
+		void RemoveLastCharacter();
+		void RemovePrevCharacter();
+		void RemoveCurrentCharacter();
+
+		void SetPosition(int position) { this->current_position = position; };
```
Alberth289346: Code in .cpp suggests our code style is to align the curly brackets at the start of the line, which in `.h` means at the `v` of `void`.

Alberth289346:
Hello sgt, long time no see,
I don't know what you aim for with this PR, but in case you want to move further I gave it a quick check. It has been too long to remember all the details though, so likely I am missing stuff.

What I didn't say was that the patch has a severe lack of doc comments, but you likely were already aware of that.

sgtbigman:
Alberth: Hello :) Thanks for giving this a quick check.

This branch definitely needs some work, so my apologies for the mess. I will address your comments and work on refactoring the overall code while I am in the process.

PR #150, opened 2021-01-21
Remove deleted rides from world

Noordfrees:
This fixes a bug with deletion of rides. Steps to reproduce the issue in master:
- Go to Buy Ride, select something, click Select
- Place the shop or coaster
- Delete it by clicking Remove in its window and confirming
- Click Select in the Ride Selection window again → the previously deleted item becomes visible.

### DIFF
--- a/src/coaster.cpp
+++ b/src/coaster.cpp
@@ -721,6 +721,22 @@ int CoasterInstance::GetFirstPlacedTrackPiece() const
 	return -1;
 }
 
+std::vector<XYZPoint16> CoasterInstance::get_all_positions() {
+	std::vector<XYZPoint16> result;
+
+	const int start_piece = GetFirstPlacedTrackPiece();
+	if (start_piece < 0) return result;
+
+	result.push_back(this->pieces[start_piece].base_voxel);
+	for (int p = start_piece;;) {
+		p = FindSuccessorPiece(this->pieces[p]);
+		if (p < 0 || p == start_piece) break;
+		result.push_back(this->pieces[p].base_voxel);
+	}
+
+	return result;
+}
+
 /**
  * Find the first placed track piece at a given position with a given entry connection.
  * @param vox Required voxel position.
--- a/src/coaster.cpp
+++ b/src/coaster.cpp
@@ -721,18 +721,16 @@ int CoasterInstance::GetFirstPlacedTrackPiece() const
 	return -1;
 }
 
-std::vector<XYZPoint16> CoasterInstance::GetAllPiecePositions() {
-	std::vector<XYZPoint16> result;
+void CoasterInstance::RemoveFromWorld()
+{
+	const uint16 index = this->GetIndex();
 	const int start_piece = GetFirstPlacedTrackPiece();
-	if (start_piece < 0) return result;
+	if (start_piece < 0) return;
 	for (int p = start_piece;;) {
-		for (const TrackVoxel *subpiece : this->pieces[p].piece->track_voxels) {
-			result.emplace_back(this->pieces[p].base_voxel + subpiece->dxyz);
-		}
+		this->pieces[p].RemoveFromWorld(index);
 		p = FindSuccessorPiece(this->pieces[p]);
 		if (p < 0 || p == start_piece) break;
 	}
-	return result;
 }
 
 /**
--- a/src/coaster.cpp
+++ b/src/coaster.cpp
@@ -721,17 +721,17 @@ int CoasterInstance::GetFirstPlacedTrackPiece() const
 	return -1;
 }
 
-std::vector<XYZPoint16> CoasterInstance::get_all_positions() {
+std::vector<XYZPoint16> CoasterInstance::GetAllPiecePositions() {
 	std::vector<XYZPoint16> result;
 
 	const int start_piece = GetFirstPlacedTrackPiece();
 	if (start_piece < 0) return result;
 
-	result.push_back(this->pieces[start_piece].base_voxel);
+	result.emplace_back(this->pieces[start_piece].base_voxel);
 	for (int p = start_piece;;) {
 		p = FindSuccessorPiece(this->pieces[p]);
 		if (p < 0 || p == start_piece) break;
-		result.push_back(this->pieces[p].base_voxel);
+		result.emplace_back(this->pieces[p].base_voxel);
 	}
 
 	return result;
--- a/src/coaster.cpp
+++ b/src/coaster.cpp
@@ -723,19 +723,14 @@ int CoasterInstance::GetFirstPlacedTrackPiece() const
 
 std::vector<XYZPoint16> CoasterInstance::GetAllPiecePositions() {
 	std::vector<XYZPoint16> result;
-
 	const int start_piece = GetFirstPlacedTrackPiece();
-	if (start_piece < 0) return result;
-
-	result.emplace_back(this->pieces[start_piece].base_voxel);
 	for (int p = start_piece;;) {
-		p = FindSuccessorPiece(this->pieces[p]);
-		if (p < 0 || p == start_piece) break;
 		for (const TrackVoxel *subpiece : this->pieces[p].piece->track_voxels) {
 			result.emplace_back(this->pieces[p].base_voxel + subpiece->dxyz);
 		}
+		p = FindSuccessorPiece(this->pieces[p]);
+		if (p < 0 || p == start_piece) break;
 	}
-
 	return result;
 }
 
--- a/src/coaster.cpp
+++ b/src/coaster.cpp
@@ -731,7 +731,9 @@ std::vector<XYZPoint16> CoasterInstance::GetAllPiecePositions() {
 	for (int p = start_piece;;) {
 		p = FindSuccessorPiece(this->pieces[p]);
 		if (p < 0 || p == start_piece) break;
-		result.emplace_back(this->pieces[p].base_voxel);
+		for (const TrackVoxel *subpiece : this->pieces[p].piece->track_voxels) {
+			result.emplace_back(this->pieces[p].base_voxel + subpiece->dxyz);
+		}
 	}
 
 	return result;
--- a/src/coaster.cpp
+++ b/src/coaster.cpp
@@ -724,6 +724,7 @@ int CoasterInstance::GetFirstPlacedTrackPiece() const
 std::vector<XYZPoint16> CoasterInstance::GetAllPiecePositions() {
 	std::vector<XYZPoint16> result;
 	const int start_piece = GetFirstPlacedTrackPiece();
+	if (start_piece < 0) return result;
 	for (int p = start_piece;;) {
 		for (const TrackVoxel *subpiece : this->pieces[p].piece->track_voxels) {
 			result.emplace_back(this->pieces[p].base_voxel + subpiece->dxyz);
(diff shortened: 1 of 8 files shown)


src/coaster.cpp
```diff
@@ -721,6 +721,22 @@ int CoasterInstance::GetFirstPlacedTrackPiece() const
 	return -1;
 }
 
+std::vector<XYZPoint16> CoasterInstance::get_all_positions() {
```
Alberth289346: This clearly lacks documentation. Should not be here but in the base class though, and since it's a pure virtual, there should be a documentation block separate from the body.
It's too long ago, but I am guessing that `RemoveAllPeople` has one too, as that function is also pure virtual, it seems.

The name of the function is in the wrong style, it should be `GetAllPositions`.

Not sure if 'positions' is the right word, `PiecePositions` comes to mind as an alternative, or `Voxels`.
Also, you should check if you got everything, afaik a piece could be bigger than one voxel.

src/coaster.cpp
```diff
@@ -721,6 +721,22 @@ int CoasterInstance::GetFirstPlacedTrackPiece() const
...
+	std::vector<XYZPoint16> result;
+
+	const int start_piece = GetFirstPlacedTrackPiece();
+	if (start_piece < 0) return result;
+
+	result.push_back(this->pieces[start_piece].base_voxel);
+	for (int p = start_piece;;) {
+		p = FindSuccessorPiece(this->pieces[p]);
+		if (p < 0 || p == start_piece) break;
+		result.push_back(this->pieces[p].base_voxel);
```
Alberth289346: `emplace_back` is more efficient, although the compiler will likely eliminate it anyway.

src/ride_type.cpp
```diff
@@ -687,6 +686,13 @@ void RidesManager::DeleteInstance(uint16 num)
 	assert(num < lengthof(this->instances));
 	this->instances[num]->RemoveAllPeople();
 	_guests.NotifyRideDeletion(this->instances[num]);
+
+	for (const XYZPoint16& position : this->instances[num]->get_all_positions()) {
+		Voxel& voxel = *_world.GetCreateVoxel(position, false);
```
Alberth289346: You just hate pointers eh?  :)

src/ride_type.h
```diff
@@ -142,6 +142,8 @@ class RideInstance {
 	virtual void RemoveAllPeople() = 0;
 	bool CanBeVisited(const XYZPoint16 &vox, TileEdge edge) const;
 
+	virtual std::vector<XYZPoint16> get_all_positions() = 0;
+
```
Alberth289346: Just insert it in a single line between 142 and 143?
It's about removal as well, just like `RemoveAllPeople`

src/coaster.cpp
```diff
@@ -721,17 +721,19 @@ int CoasterInstance::GetFirstPlacedTrackPiece() const
...
 
-	result.push_back(this->pieces[start_piece].base_voxel);
+	result.emplace_back(this->pieces[start_piece].base_voxel);
 	for (int p = start_piece;;) {
 		p = FindSuccessorPiece(this->pieces[p]);
 		if (p < 0 || p == start_piece) break;
-		result.push_back(this->pieces[p].base_voxel);
+		for (const TrackVoxel *subpiece : this->pieces[p].piece->track_voxels) {
+			result.emplace_back(this->pieces[p].base_voxel + subpiece->dxyz);
+		}
```
Alberth289346: Sorry, but after a night of sleep this bothered me. It looks like the wrong approach to me, since you're breaking encapsulation.

A `Coaster` has `Trackpiece`s, and a `CoasterInstance` has `PositionedTrackPiece`s. Those are the classes that know internals of track pieces. Here you export that knowledge to `CoasterInstance` level.

It may look like a simple solution, but now, track piece internals are not just in track pieces, they are also here. So if you change track piece internals say one year after now, you have to change two places now. There is a large chance you are not going to keep those two spots consistent.

src/ride_type.cpp
```diff
@@ -687,10 +692,13 @@ void RidesManager::DeleteInstance(uint16 num)
...
-		Voxel& voxel = *_world.GetCreateVoxel(position, false);
-		assert(voxel.instance == num + SRI_FULL_RIDES);
-		voxel.ClearInstances();
+	for (const XYZPoint16& position : this->instances[num]->GetAllPiecePositions()) {
+		Voxel *voxel = _world.GetCreateVoxel(position, false);
+		assert(voxel);
+		if (voxel->instance != SRI_FREE) {
+			assert(voxel->instance == num + SRI_FULL_RIDES);
+			voxel->ClearInstances();
+		}
```
Alberth289346: A `RidesManager` deals with a ride at global level, high level stuff. How muich money does it make etc.
For some reason, you decided that it must now deal with trackpiece positions, something internal to the ride. I don't know why you did that.

In a real world comparison, what you're doing is giving the CEO of an amusement park a list of track pieces he should remove. Th CEO must climb into the ride, and remove each piece by him/her self.

So what you should do instead is follow the chain of command here. A CEO tells the ride owner to remove the ride, the ride owner tells the work people to remove all track pierces, and the work people figure out which voxels they have the change.
So here, you'd just say `this->instances[num]->RemoveFromWorld();`  this is what a CEO can handle comfortably.
Inside the Ride instance, you then pass it on to all parts of the ride, which then actually change voxels.

In that way, you keep internals at the level where they belong, and you're not duplicating code to other parts of the code base that eventually will get inconsistent.

src/ride_type.cpp
```diff
@@ -221,6 +221,11 @@ const RideType *RideInstance::GetRideType() const
  * Immediately remove all guests and staff which is inside the ride.
  */
 
+/**
+ * \fn void RideInstance::GetAllPiecePositions()
+ * Returns a vector with the coordinates of all voxels which are occupied by this ride.
+ */
```
Alberth289346, line 227: Your documentation is at the wrong abstraction level.

You are describing what the code is quite literally doing. That isn't much help, since you can read code to figure that out.
What you should write instead is the higher level intent of the function.
Perhaps something like "collect all voxels used by the ride".

vectors and coordinates are the irrelevant low level details, the shape of how you represent a collection of voxels here.

You may not realize it, but in this program, documentation is leading above code. The documentation described the high level intent of the function, the code realizes that intent in some form. If documentation and code are in conflict, intent wins.

The reason is that you read the documentation when performing a function call. You don't go in reading 150 lines of dense code to figure out how to call the thing. In other words, all calls to a function are based on the written intent and not on what the code is actually doing. Making code follow intent makes code fixes more local.

Noordfrees:
So, if I understand correctly, the hierarchy is like this: The RideManager owns many coasters. Each coaster owns many PositionedTrackPieces. Each PositionedTrackPiece owns one TrackPiece. Each TrackPiece occupies several voxels. So I now changed my code to pass the function call down through the various levels and let the bottommost class handle the actual cleanup. Correct?

Alberth289346:
`RideManager` is wider than coasters (I think, class documentation should state its goals). It contains any ride, including shops, gentle rides, etc. It's basically a place to keep track of what rides exist in the park, and provides (or should provide) the common interface like open/close, querying money stuff for the UI, etc.

Coasters exist in two forms. There is coaster as the type of ride (ie "steel rollercoaster") which is a collection of track pieces that the game provides for the type of coaster. A track piece is a stretch of track of basically any form, anything from a 1 tile straight track or platform to a large half-loop piece, but also eg foto-shoot. Each track piece has a base voxel (or at least a base position, don't remember it exactly), and a set of voxels relative to that base voxel. It defines the space that you need for the track piece in order to not have collisions with other rides. A track piece also has connection points relative to the base voxel so you can connect different track pieces to each other when you build a real coaster (below).
These are thus libraries of parts available for use when you build a coaster.

The second form is a coaster that you can ride, a CoasterInstance. It has a coaster type (eg "wooden rollercoaster"), which defines the set of track pieces that can be used for the track of the coaster from the library. CoasterInstance has PositionedTrackpieces, which represents trackpieces that exist in the world. A positioned track piece defines the position of the base voxel, and points at a track piece from the coaster library, thus fully defining the set voxels and connection point positions for that part of the track of the coaster instance.
Somewhere near opening a coaster there is code that verifies that all pieces together from an infinite loop without gaps and without positioned track pieces that aren't used and without connections that have no neighbour positioned track piece. It then shuffles the positioned track pieces (iirc) so a ride progresses nicely sequentially thought the positioned pieces.

As looking for occupied voxels in the world by digging through the positioned track pieces of all coaster instances is likely too slow (not too mention complicated to draw in its current form), voxels in the world point back to the coaster instance (and likely the positioned track piece, I don't remember exactly).

Alberth289346:
There is one thing that isn't optimal yet, but I have asked for so many revisions already, not sure if it's appropriate to ask this.

The problem is as follows:
``` cpp
void ShopInstance::RemoveFromWorld()
{
	const uint16 index = this->GetIndex();
	Voxel *voxel = _world.GetCreateVoxel(vox_pos, false);
	assert(voxel);
	if (voxel->instance != SRI_FREE) {
```
I grabbed an example from the shop, but it happens more often.

The `GetCreateVoxel` function tries to get a voxel, and if it doesn't exit, it creates one (as stated in the documentation of that function). Thus your `assert` will never fire in its current form.
Secondly, as your objective is to remove a ride, creating an empty voxel to check if it's empty makes very little sense.

So I would propose to change it to
``` cpp
        Voxel *voxel = _world.GetVoxel(vox_pos);
	if (voxel != nullptr && voxel->instance != SRI_FREE) {
```
I am not sure `voxel == nullptr` will ever happen. (You're supposedly using data that specifies where the ride is in the world, and thus it should be there, or it's a bug), so perhaps you can or maybe even should live without the `nullptr` check.

Noordfrees:
> There is one thing that isn't optimal yet, but I have asked for so many revisions already, not sure if it's appropriate to ask this.

Of course you may ask, I'm learning a lot about the code base as I'm addressing your comments :) 

The reason why I use `GetCreateVoxel` rather than `GetVoxel` is that the latter returns a `const Voxel*` but I need a `volatile Voxel*` here. 
Implementing a function `volatile Voxel* VoxelWorld::GetVoxel()` would require a `const_cast` which I have always been told is really bad style and should be avoided. 
According to the doc of `GetCreateVoxel` this only creates the voxel though if it does not yet exist and the second argument is `true`, so I assume that `GetCreateVoxel(x,false)` can be supposed the non-`const` equivalent of `GetVoxel(x)` (and it *can* return nullptr then so the assert is still possible).

So, would it be better if I add a non-`const` `GetVoxel()` function to `VoxelWorld` with a `const_cast` (or something like `Voxel* GetExistingVoxel()` and assert the voxel's existance there), or stick with the current code?

Alberth289346:
Const cast is indeed something you only use when there is no other way out, so if you can avoid it, please do :)

> According to the doc of GetCreateVoxel this only creates the voxel though if it does not yet exist and the second argument is true, so I assume that GetCreateVoxel(x,false)

Hmm, my memory of the code is rusty :)   Ok, it's good then.

> and it can return nullptr then so the assert is still possible

In the general case it can indeed, but as I explained you're using an administration that should point at non-empty voxels with data. In such a situation I typically don't bother with an assert, and simply let it crash on the first dereference. To avoid hunting bugs that shouldn't exist it may be useful then, to state something like `// Voxel should exist and be non-empty.` in the source, to remind the reader that the `null` option was consciously discarded as not possible :) 

Of course, the `assert` states the same thing, except in code rather than in comment.

Alberth289346:
And congrats with your first patch :)

Noordfrees:
The first of many I hope ;) 
Thank you for the review and your patience :)

PR #153, opened 2021-01-22
Shops Height

Noordfrees:
Previously all shops had a height of 1, allowing players to build paths directly above them. Now any height is supported, and the default shops are given heights of 2 (snacks, toilet) or 3 (ice) to ensure they get sufficient vertical space.

### DIFF
--- a/src/coaster.h
+++ b/src/coaster.h
@@ -244,7 +244,7 @@ public:
 
 	void RemoveFromWorld() override;
 	void InsertIntoWorld() override {
-		// This was already done during construction – nothing left to do here
+		/* This was already done during construction – nothing left to do here. */
 	}
 
 	PositionedTrackPiece *pieces; ///< Positioned track pieces.
--- a/src/coaster.h
+++ b/src/coaster.h
@@ -243,6 +243,9 @@ public:
 	void Save(Saver &svr);
 
 	void RemoveFromWorld() override;
+	void InsertIntoWorld() override {
+		// This was already done during construction – nothing left to do here
+	}
 
 	PositionedTrackPiece *pieces; ///< Positioned track pieces.
 	int capacity;                 ///< Number of entries in the #pieces.
--- a/src/ride_build.cpp
+++ b/src/ride_build.cpp
@@ -178,32 +178,25 @@ void RideBuildWindow::OnClick(WidgetNumber wid_num, const Point16 &pos)
  */
 bool RideBuildWindow::CanPlaceShop(const ShopType *selected_shop, const XYZPoint16 &pos, ViewOrientation vp_orient)
 {
-	// This function assumes that all shop have size 1×1.
+	// This function assumes that all shops have size 1×1.
 
 	/* 1. Can the position itself be used to build a shop? */
 	if (_world.GetTileOwner(pos.x, pos.y) != OWN_PARK) return false;
-	const Voxel *vx = _world.GetVoxel(pos);
-	if (vx != nullptr) {
+
+	auto check_if_location_is_suited = [&pos, selected_shop](const int8 offset, const bool need_flat_ground) {
+		const Voxel *vx = _world.GetVoxel(pos + XYZPoint16(0, 0, offset));
+		if (!vx || vx->GetGroundType() == GTP_INVALID || (need_flat_ground != (vx->GetGroundSlope() == SL_FLAT))) return false;
 		for (int8 h = selected_shop->height - 1; h >= 0; --h) {
-			const Voxel *v = _world.GetVoxel(pos + XYZPoint16(0, 0, h));
-			if (v && !v->CanPlaceInstance()) return false; // Cannot build on a path or other ride.
-		}
-
-		if (vx->GetGroundType() != GTP_INVALID && vx->GetGroundSlope() == SL_FLAT) return true; // Can build at a flat surface.
-	}
-
-	/* 2. Is the shop just above non-flat ground? */
-	if (pos.z > 0) {
-		vx = _world.GetVoxel(pos + XYZPoint16(0, 0, -1));
-
-		for (int8 h = selected_shop->height - 1; h >= 0; --h) {
-			const Voxel *v = _world.GetVoxel(pos + XYZPoint16(0, 0, h - 1));
+			const Voxel *v = _world.GetVoxel(pos + XYZPoint16(0, 0, h + offset));
 			if (v && !v->CanPlaceInstance()) return false;
 		}
+		return true;
+	};
 
-		if (vx != nullptr && vx->CanPlaceInstance() &&
-				vx->GetGroundType() != GTP_INVALID && vx->GetGroundSlope() != SL_FLAT) return true;
-	}
+	if (check_if_location_is_suited(0, true)) return true;
+
+	/* 2. Is the shop just above non-flat ground? */
+	if (pos.z > 0 && check_if_location_is_suited(-1, false)) return true;
 
 	/* 3. Is there a path at the right place? */
 	for (TileEdge entrance = EDGE_BEGIN; entrance < EDGE_COUNT; entrance++) { // Loop over the 4 unrotated directions.
--- a/src/ride_build.cpp
+++ b/src/ride_build.cpp
@@ -178,21 +178,19 @@ void RideBuildWindow::OnClick(WidgetNumber wid_num, const Point16 &pos)
  */
 bool RideBuildWindow::CanPlaceShop(const ShopType *selected_shop, const XYZPoint16 &pos, ViewOrientation vp_orient)
 {
-	// This function assumes that all shops have size 1×1.
+	/* This function assumes that all shops have size 1×1. */
 
 	/* 1. Can the position itself be used to build a shop? */
 	if (_world.GetTileOwner(pos.x, pos.y) != OWN_PARK) return false;
-
 	auto check_if_location_is_suited = [&pos, selected_shop](const int8 offset, const bool need_flat_ground) {
 		const Voxel *vx = _world.GetVoxel(pos + XYZPoint16(0, 0, offset));
-		if (!vx || vx->GetGroundType() == GTP_INVALID || (need_flat_ground != (vx->GetGroundSlope() == SL_FLAT))) return false;
+		if (vx == nullptr || vx->GetGroundType() == GTP_INVALID || (need_flat_ground != (vx->GetGroundSlope() == SL_FLAT))) return false;
 		for (int8 h = selected_shop->height - 1; h >= 0; --h) {
 			const Voxel *v = _world.GetVoxel(pos + XYZPoint16(0, 0, h + offset));
-			if (v && !v->CanPlaceInstance()) return false;
+			if (v != nullptr && !v->CanPlaceInstance()) return false;
 		}
 		return true;
 	};
-
 	if (check_if_location_is_suited(0, true)) return true;
 
 	/* 2. Is the shop just above non-flat ground? */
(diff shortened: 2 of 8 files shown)


Alberth289346:
I had a quick look, and you got the idea I had when shops were added. I am just not sure this is still the right approach.

Shops were added after roads, so it was the first thing that wasn't completely flat. At the same time, I had no idea about complexities involved in roller coasters. As a result the current solution was born.

In the mean time, roller coasters were added, with track pieces. Also, missing currently is multi-tile buildings, eg cinema but also the ship and other attractions that move its customers in all possible directions.

So the first thought I had when I read your patch was, "is the 1 tile shop worthwhile to have a special case for?", in particular, can't it be a track-piece like thing that defines its own space? (ie a track piece can also have empty space, so why code "height" explicitly here?)
Also, can't a shop be a multi-tile building that just happens to be a bit small? The argument against it are the entrance and exit buildings, but that may be solvable, eg by having several forms of connection points??? [Just thinking aloud here, I have no answers currently]

The other 1 tile "building" that we have are the roads. Currently they have an implicit fixed height I guess. I wonder about that too. They'd fit in a track-piece like definition too, but there is never a need to change their height (or at the very least height depends on the visitors rather than the road itself).
So roads seem in favour of having a height concept, which in turn makes having shops a height not so bad either.

Of course you can have "height" in the data file, and right after loading convert it to a track-piece like thing that defines the set of voxels that it needs.

What I can't quite decide is how a multi-tile building should work. Would it be a track-piece like thing, so the designer has freedom in defining the voxels that it needs, or does it have a height for every tile, or just one height?

And closer to the patch, should a shop be a thing of its own, or be a small multi-tile building?


Another thing I don't understand in your patch is the need to have `InsertInWorld`. ie the shop can already be build, so why have another 'build' function?  Didn't spend much time to understand your patch though, so it may become clear when I do that.

Having a nice walk now, I'll have a closer look later today.

Noordfrees:
How about introducing a hierarchy
```
class FixedRideType  : public RideType
class ShopType       : public FixedRideType
class GentleRideType : public FixedRideType
```
(with corresponding `*Instance` classes)?

Then all the stuff that is now shop-specific can be moved to the abstract `FixedRideType` and `FixedRideInstance` to be reused later for cinemas etc. `FixedRideType` could have properties `width_x`, `width_y`, and `height[width_x*width_y]` to allow the ride designer to allow most any shape they like.

src/ride_type.cpp
```diff
@@ -226,6 +226,23 @@ const RideType *RideInstance::GetRideType() const
  * Immediately remove this ride from all voxels it currently occupies.
  */
 
+/**
+ * \fn void RideInstance::InsertIntoWorld()
+ * Ensure that this shop is linked into the voxels it is meant to occupy.
+ */
+
+/**
+ * \fn void ShouldDrawPiece::InsertIntoWorld()
```
Alberth289346: Class and function name are the wrong way around, but since you write this just above the function, you don't need this line at all, doxygen understands the text belongs to the next function by default

src/ride_type.cpp
```diff
@@ -226,6 +226,23 @@ const RideType *RideInstance::GetRideType() const
...
+ * Whether the specified piece of the ride should be visible to the player.
+ * @param voxel_number The instance data of the piece to check.
+ * @return `false` if the ride piece is meant to be fully invisible; `true` otherwise.
```
Alberth289346: I typically write that as a statement like you wrote at line 236. Iff the answer to the statement is true you must return `true`.

Note what you wrote is not wrong, it also makes clear what a returned value means.
Doxygen however doesn't use markdown syntax, your 'false' and 'true' wont be console style. Use "\c false" (without quotes) to get that.

src/ride_type.cpp
```diff
@@ -226,6 +226,23 @@ const RideType *RideInstance::GetRideType() const
...
+ */
+bool RideInstance::ShouldDrawPiece(uint16 voxel_number) const
+{
+	// By default, all pieces are drawn, unless a subclass implements a special case
```
Alberth289346: Comments are normal sentences, they start with an upper case letter (which you do), and end with a dot.

src/ride_build.cpp
```diff
@@ -178,18 +178,30 @@ void RideBuildWindow::OnClick(WidgetNumber wid_num, const Point16 &pos)
...
 		vx = _world.GetVoxel(pos + XYZPoint16(0, 0, -1));
-		if (vx != nullptr && vx->GetInstance() == SRI_FREE &&
+
+		for (int8 h = selected_shop->height - 1; h >= 0; --h) {
+			const Voxel *v = _world.GetVoxel(pos + XYZPoint16(0, 0, h - 1));
+			if (v && !v->CanPlaceInstance()) return false;
+		}
+
+		if (vx != nullptr && vx->CanPlaceInstance() &&
 				vx->GetGroundType() != GTP_INVALID && vx->GetGroundSlope() != SL_FLAT) return true;
```
Alberth289346: Don't duplicate code. Make a local function and call that twice.

src/ride_type.cpp
```diff
@@ -631,6 +648,8 @@ void RidesManager::NewInstanceAdded(uint16 num)
 	const RideType *rt = ri->GetRideType();
 	assert(ri->state == RIS_ALLOCATED);
 
+	ri->InsertIntoWorld();
+
```
Alberth289346, line 639: I think we should discuss empty lines.

You seem to like double-spaced code (one line code one line empty) for some reason, but it throws away the value of an empty line.

In the program, code that perform a 'step' (a closely releated coherent set of actions towards the final goal) has no empty lines. There is an empty between steps. Often a step begins with a comment line explaining what the step is aiming to achieve (this is not a rule, but it's useful for browsing code).

What the above does is making visually separated blocks of code that belong together. Especially with comment lines above blocks, you can simply skip blocks of code by just reading the comment line, until you found the block that tries to do what you're looking for. Only at that point you have to start reading actual code.

If you make single line "blocks" this system breaks, you still have to read each line of code, at which point it's better to simply delete all empty lines, as it shrinks the function by 50% lines.

So please avoid making 1-line blocks, probably by making the 'step' that a block performs bigger or at a higher abstraction level.

In this case, adding the new line to the block above seems a good solution.

src/shop_type.cpp
```diff
@@ -281,14 +303,18 @@ void ShopInstance::Load(Loader &ldr)
...
-	} else {
-		ldr.SetFailMessage("Invalid world coordinates for shop.");
+	const int16 height = this->GetShopType()->height;
+	for (int16 i = 0; i < height; ++i) {
+		Voxel *v = _world.GetCreateVoxel(this->vox_pos + XYZPoint16(0, 0, i), true);
+		if (v != nullptr && v->GetInstance() == SRI_FREE) {
+			v->SetInstance(inst_number);
+			v->SetInstanceData(i > 0 ? INVISIBLE_SHOP_TILE : entrances);
+
+			AddRemovePathEdges(this->vox_pos, PATH_EMPTY, entrances, PAS_QUEUE_PATH);
```
Alberth289346: You're running this more than once now (if `height > 1`), but it doesn't seem to depend on loop data `i` or `v`.

Alberth289346:
I can see the value of `InsertIntoWorld`, but I am not yet convinced by `ShouldDrawPiece`. How does a rollercoaster not paint all its claimed voxels? (The simplest way would be to claim the tile but not have any sprite to paint in the tile. Don't know if that's possible currently.)

I also am not sure a voxel is small enough, eg garden parts are 1/4 tile in the original game. Don't know if a voxel supports that curently though.

Alberth289346:
> Then all the stuff that is now shop-specific can be moved to the abstract FixedRideType and FixedRideInstance to be reused later for cinemas etc. FixedRideType could have properties width_x, width_y, and height[width_x*width_y] to allow the ride designer to allow most any shape they like.

I wasn't thinking in how to implement the answer yet, just pondering what the right answer would be, before considering how to implement it.

I do that for more complicated questions to avoid "correct answer" == "easy to implement" connections, which are confusing, and possibly harmful if the easy to implement answer turns out to be not quite correct.

In this case, your idea is probably fine, the only counter-reason I could think of is that with your idea a shop or multi-tile ride is different from a track piece, but that is likely not a big problem at worst case.

EDIT: So let's keep height, and let's add an additional class layer when multi-tile rides arrive.

src/ride_build.cpp
Noordfrees: ok done

src/shop_type.cpp
Noordfrees: fixed

Noordfrees:
Getting a shop (or any other ride) not to render anything onto a given voxel is easy enough to do (just set `sprites[1] = nullptr` in `…Instance::GetSprites`), but the reason the `ShouldDrawPiece` business was needed was that a platform was rendered at all height steps regardless. But I found a better solution for that now and removed `ShouldDrawPiece` again.

src/ride_type.cpp
Noordfrees: Okay. I'm still getting accustomed to the code style used in this project, but I'm confident such mistakes will get less frequent quickly.

src/ride_build.cpp
```diff
@@ -178,20 +178,25 @@ void RideBuildWindow::OnClick(WidgetNumber wid_num, const Point16 &pos)
...
 	if (_world.GetTileOwner(pos.x, pos.y) != OWN_PARK) return false;
-	const Voxel *vx = _world.GetVoxel(pos);
-	if (vx != nullptr) {
-		if (!vx->CanPlaceInstance()) return false; // Cannot build on a path or other ride.
-		return vx->GetGroundType() != GTP_INVALID && vx->GetGroundSlope() == SL_FLAT; // Can build at a flat surface.
-	}
+
+	auto check_if_location_is_suited = [&pos, selected_shop](const int8 offset, const bool need_flat_ground) {
+		const Voxel *vx = _world.GetVoxel(pos + XYZPoint16(0, 0, offset));
+		if (!vx || vx->GetGroundType() == GTP_INVALID || (need_flat_ground != (vx->GetGroundSlope() == SL_FLAT))) return false;
```
Alberth289346: Code doesn't negate pointers, it always checks against `nullptr` constant.

src/coaster.h
```diff
@@ -243,6 +243,9 @@ class CoasterInstance : public RideInstance {
 	void Save(Saver &svr);
 
 	void RemoveFromWorld() override;
+	void InsertIntoWorld() override {
+		// This was already done during construction – nothing left to do here
```
Alberth289346: Sentence needs a dot at the end.

Also, I think the code style says to use C style comment here (`/* .. */`), `//` is only after code.
I am not too fussed about it though.

src/ride_build.cpp
```diff
@@ -178,20 +178,25 @@ void RideBuildWindow::OnClick(WidgetNumber wid_num, const Point16 &pos)
...
+		for (int8 h = selected_shop->height - 1; h >= 0; --h) {
+			const Voxel *v = _world.GetVoxel(pos + XYZPoint16(0, 0, h + offset));
+			if (v && !v->CanPlaceInstance()) return false;
```
Alberth289346: also here check against `!= nullptr`

Alberth289346:
In `TileEdge` (tile.h line 146), there is `INVALID_EDGE`, which looks like a better alternative to the `INVISIBLE_SHOP_TILE` constant.

EDIT: The instance data should be a bitset I guess, or you cannot have several directions. In that case, `0` would be the right alternative, although I'd expect a constant for it as well.

EDIT: It is a bitset, see `ShopFlags` (ride_type.h, line 42)

PR #157, opened 2021-01-24
Fix assert fail when closing Ride Selection window

Noordfrees:
The bug can be reproduced in master by opening the Ride Selection window, choosing a shop, clicking Select, and then closing the Ride Selection window while the Place Shop window is still open.

I wondered whether the assert might be guarding against a possible memory leak or heap-use-after-free, so I compiled&linked with ASan and it turns out that this is not the case, so it can safely be removed.

### DIFF
--- a/src/ride_gui.cpp
+++ b/src/ride_gui.cpp
@@ -129,10 +129,7 @@ RideSelectGui::RideSelectGui() : GuiWindow(WC_RIDE_SELECT, ALL_WINDOWS_OF_TYPE)
 	this->SetNewRideKind(this->current_kind, true);
 }
 
-RideSelectGui::~RideSelectGui()
-{
-	_rides_manager.CheckNoAllocatedRides(); // Check that no rides are being constructed.
-}
+RideSelectGui::~RideSelectGui() = default;
 
 void RideSelectGui::UpdateWidgetSize(WidgetNumber wid_num, BaseWidget *wid)
 {
@@ -244,6 +241,7 @@ void RideSelectGui::OnClick(WidgetNumber wid_num, const Point16 &pos)
 					_rides_manager.NewInstanceAdded(instance);
 					ShowCoasterManagementGui(ri);
 				}
+				delete this;
 			}
 			break;
 		}
--- a/src/ride_gui.cpp
+++ b/src/ride_gui.cpp
@@ -26,7 +26,7 @@
 class RideSelectGui : public GuiWindow {
 public:
 	RideSelectGui();
-	~RideSelectGui();
+	~RideSelectGui() = default;
 
 	void UpdateWidgetSize(WidgetNumber wid_num, BaseWidget *wid) override;
 	void DrawWidget(WidgetNumber wid_num, const BaseWidget *wid) const override;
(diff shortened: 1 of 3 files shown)


src/ride_gui.cpp
```diff
@@ -129,10 +129,7 @@ RideSelectGui::RideSelectGui() : GuiWindow(WC_RIDE_SELECT, ALL_WINDOWS_OF_TYPE)
 	this->SetNewRideKind(this->current_kind, true);
 }
 
-RideSelectGui::~RideSelectGui()
-{
-	_rides_manager.CheckNoAllocatedRides(); // Check that no rides are being constructed.
-}
+RideSelectGui::~RideSelectGui() = default;
```
Alberth289346: Just remove this line and the `~RideSelectGui()` in the class declaration? (Or set it to `default` there? not sure if there is any difference, I'd expect not.)

src/ride_gui.cpp
```diff
@@ -244,6 +241,7 @@ void RideSelectGui::OnClick(WidgetNumber wid_num, const Point16 &pos)
 					_rides_manager.NewInstanceAdded(instance);
 					ShowCoasterManagementGui(ri);
 				}
+				delete this;
```
Alberth289346: This has nothing to do with fixing the assert as far as I can see, so it doesn't quite belong in this commit.

Is there a rationale why removing the selection window after making a selection improves UX?

Noordfrees: It slipped in by accident, though I'd still like to propose this change when it's already there. I believe most of the time, people build one ride, then take their time to set up paths and queues, change the ride's pricing, colouring and settings, etc. For tracked rides, they additionally spend a lot of time designing the track and running the ride in test mode first. Most of this is not implemented yet, but the Ride Selection window is not used again for quite some time after settling on which ride to construct, and it only blocks the view on the map unnecessarily. So I would prefer that this window auto-closes after the user hits Select. I don't feel strongly about this though.

src/ride_gui.cpp
Noordfrees: Moved it to the class declaration. I've heard some compilers can optimize better if ctors and dtors are explicitly defaulted than if they're missing or have an empty body.

src/ride_gui.cpp
```diff
@@ -26,7 +26,7 @@
 class RideSelectGui : public GuiWindow {
 public:
 	RideSelectGui();
-	~RideSelectGui();
+	~RideSelectGui() = default;
```
Alberth289346: Just for my curiosity, is this different from simply deleting the line entirely?
(If not, why keep the line?)

src/ride_gui.cpp
Alberth289346: The point is, if you read the list commit messages, it doesn't say "auto-close ride selection window" ever, so finding the commit that changes the behavior becomes a search party, probably with bi-section of by searching in git for changes on that line.

I have no string feeling about either, so if you want it in, that's fine. Please make it a separate commit, to enhance chances of finding it again if the need arises.

src/ride_gui.cpp
Noordfrees: I've heard that some compilers can optimize better if ctors and dtors are explicitly defaulted than if they have an empty body or are missing. (I can't think of examples right now though, so if you prefer I could also remove the line.)

src/ride_gui.cpp
Noordfrees: moved to #159

src/ride_gui.cpp
Noordfrees: I checked this again and apparently only empty dtor bodies are a problem for optimizing. Removed the line.

PR #158, opened 2021-01-24
Forbid building shops inside slopes

Noordfrees:
This fixes the following bug:
grafik

### DIFF
--- a/src/ride_build.cpp
+++ b/src/ride_build.cpp
@@ -168,6 +168,48 @@ void RideBuildWindow::OnClick(WidgetNumber wid_num, const Point16 &pos)
 }
 
 /**
+ * Checks whether the air space above the ground at the given location is suited to place a shop of the given height.
+ * @param position Coordinate of the base voxel.
+ * @param height Height of the object.
+ * @return The space is suited to build the shop.
+ */
+static bool CheckSufficientVerticalSpace(const XYZPoint16& position, const int8 height)
+{
+	for (int8 h = 0; h <= height; ++h) {
+		const Voxel *v = _world.GetVoxel(position + XYZPoint16(0, 0, h));
+		if (v != nullptr && !v->CanPlaceInstance()) return false;
+	}
+	return true;
+}
+
+/**
+ * Checks whether the given location is suited to place a shop of the given height on flat ground.
+ * @param position Coordinate of the base voxel.
+ * @param height Height of the object.
+ * @return The space is flat and suited to build the shop.
+ */
+static bool CanPlaceShopOnFlatGround(const XYZPoint16& position, const int8 height)
+{
+	const Voxel *vx = _world.GetVoxel(position);
+	return vx != nullptr && vx->GetGroundType() != GTP_INVALID && vx->GetGroundSlope() == SL_FLAT && CheckSufficientVerticalSpace(position, height);
+}
+
+/**
+ * Checks whether the given location is suited to place a shop of the given height on a slope.
+ * @param position Coordinate of the base voxel.
+ * @param height Height of the object.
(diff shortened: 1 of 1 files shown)


src/ride_build.cpp
```diff
@@ -182,19 +182,33 @@ bool RideBuildWindow::CanPlaceShop(const ShopType *selected_shop, const XYZPoint
...
+			if (vx->GetGroundSlope() != SL_FLAT) return false;
+		} else {
+			if (vx->GetGroundSlope() == SL_FLAT) return false;
+			const Voxel *top_voxel = _world.GetVoxel(pos);
+			if (top_voxel != nullptr) {
+				const uint8 slope = top_voxel->GetGroundSlope();
+				if (slope & ISL_BOTTOM_STEEP_NORTH) return false;
+				if (slope & ISL_BOTTOM_STEEP_SOUTH) return false;
+				if (slope & ISL_BOTTOM_STEEP_EAST) return false;
+				if (slope & ISL_BOTTOM_STEEP_WEST) return false;
```
Alberth289346: This is definitely wrong. `ISL_BOTTOM_STEEP_*` is an imploded slope description (ie enumeration values that define sprite numbers) see tile.h, line 81.
While you can do bit operations on two random integer numbers, it's not likely to do much good.

You can expand the slope to a collection of bits using something like `TileSlope ts = ExpandTileSlope(v->GetGroundSlope());`, `TileSlope` as well as the expand function are in tile.h.

Alternatively, use `IsImplodedSteepSlope(slope)` also in tile.h. A third option is to move the shop one tile higher, on top of the steep slope.

As for code style, here you should compare against not being 0. Only boolean values don't need a comparison.

and last but not least, if you want to test for at least one of several bits, you can merge the comparison with `(v & (B1 | B2 | ... | Bn) != 0)` for bit patterns `B1`, `B2`, ..., `Bn`.

src/ride_build.cpp
```diff
@@ -182,19 +182,33 @@ bool RideBuildWindow::CanPlaceShop(const ShopType *selected_shop, const XYZPoint
 
 	/* 1. Can the position itself be used to build a shop? */
 	if (_world.GetTileOwner(pos.x, pos.y) != OWN_PARK) return false;
-	auto check_if_location_is_suited = [&pos, selected_shop](const int8 offset, const bool need_flat_ground) {
+	auto check_if_location_is_suited = [&pos, selected_shop](const bool is_flat_ground) {
```
Alberth289346: The lambda function explodes somewhat in size.

Maybe make a new normal function for it? I don't know what Doxygen does with this, but I am guessing it won't get listed as a function. This is a shame, as `check_if_location_is_suited` is a quite nice abstraction, that may be useful at other places.

src/ride_build.cpp
```diff
@@ -182,19 +182,33 @@ bool RideBuildWindow::CanPlaceShop(const ShopType *selected_shop, const XYZPoint
...
 			if (v != nullptr && !v->CanPlaceInstance()) return false;
 		}
 		return true;
 	};
-	if (check_if_location_is_suited(0, true)) return true;
+	if (check_if_location_is_suited(true)) return true;
 
 	/* 2. Is the shop just above non-flat ground? */
-	if (pos.z > 0 && check_if_location_is_suited(-1, false)) return true;
+	if (pos.z > 0 && check_if_location_is_suited(false)) return true;
```
Alberth289346: I don't have an answer here, but these `true` and `false` magic values always bother me.

The only option I can see is to split the function into 2 functions, so you can express the meaning of the boolean in the function name. Not sure that is worthwhile though.

PR #161, opened 2021-01-27
Gentle and Thrill Rides

Noordfrees:
Gentle rides and thrill rides are handled by new classes `GentleThrillRideType/Instance`. Since these two types of rides behave exactly the same there's no need to have two separate class structures. They are distinguished only by the ride type (`RTK_GENTLE` or `RTK_THRILL`).

A lot of common code for shops and gentle/thrill rides was pulled out into a new abstract class layer `FixedRideType/Instance`.

I created one gentle ride meant to be based on RTC's Haunted House with placeholder graphics.

grafik


Alberth289346:
Wow, you're fast :)
Looks good at first sight, will check in more detail tomorrow.

doc/data_format.rst
```diff
@@ -596,6 +596,47 @@ Version history
...
+======  ======  =======  ===================================================================================
+   0       4      1-     Magic string 'FGTR'.
+   4       4      1-     Version number of the block.
+   8       4      1-     Length of the block excluding magic string, version, and length.
+  12       1      1-     Magic number 1 for thrill rides or 0 for gentle rides.
+  13       2      1-     Zoom-width of a tile of the surface.
+  15       1      1-     Length of the ride in x direction in voxels.
+  16       1      1-     Length of the ride in y direction in voxels.
+  17       x      1-     Heights of the ride in voxels. The number x of height bytes is
+                         equal to the product of the ride lengths in x and y direction.
```
Alberth289346: Having `x` and `y` directions, as well as an `x` in the length column of this field is confusing. Instead of `x` use a different letter to avoid it, for example `s` (size) or `c` (count) but any other letter will do except `x` and `y`.

The explanation is a bit short, perhaps `Heights of the ride in voxels, for each tile occupied by the ride.` ? (ie also explain why you need `x` bytes).

Explanation about value `x` can be added here or below the table.

doc/data_format.rst
```diff
@@ -596,6 +596,47 @@ Version history
...
+  29+17*x  4      1-     nw preview.
+  33+17*x  4      1-     First recolouring specification.
+  37+17*x  4      1-     Second recolouring specification.
+  41+17*x  4      1-     Third recolouring specification.
+  45+17*x  4      1-     Entrance fee.
+  49+17*x  4      1-     Monthly cost of having the ride.
+  53+17*x  4      1-     Additional monthly cost of having an opened ride.
+  57+17*x  4      1-     Text of the ride (reference to a TEXT block).
+  61+17*x                Total length.
+
```
Alberth289346: No empty line.

Also, the `=====` blocks must be at least as wide as the text in the column. (restructured text is quite precise here.)

Unfortunately, Markdown doesn't have standard table support. Perhaps eventually something like Asciidoc would work, but I would say there are more interesting things to do than reformatting documentation.

doc/data_format.rst
```diff
@@ -596,6 +596,47 @@ Version history
...
+  33+17*x  4      1-     First recolouring specification.
+  37+17*x  4      1-     Second recolouring specification.
+  41+17*x  4      1-     Third recolouring specification.
+  45+17*x  4      1-     Entrance fee.
+  49+17*x  4      1-     Monthly cost of having the ride.
+  53+17*x  4      1-     Additional monthly cost of having an opened ride.
+  57+17*x  4      1-     Text of the ride (reference to a TEXT block).
+  61+17*x                Total length.
+  
+======  ======  =======  ===================================================================================
```
Alberth289346: The thing I was wondering about is animation.

There is a preview, and a single full display of the building in 4 orientations. That works for shops, cinemas, and other closed buildings, but not for thrill rides with moving parts.

doc/data_format.rst
```diff
@@ -596,6 +596,47 @@ Version history
 - 5 (20141010) Added more items.
 
 
+Fixed gentle and thrill rides
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Gentle and thrill rides consisting of a single building. FreeRCT can read block version 1.
```
Alberth289346, line 654: What does `Fixed` mean? Line 601 speaks about "single building", but doesn't connect that to "fixed".

If you introduce a new word with a specific intended meaning, explain it so everybody has the same understanding what the word means. It avoids situations where several people talk about 'fixed' but have different notions of what it means.

Currently, one can read 'fixed' as 'not broken', as 'lack of animation', as 'not changing in size' or as 'single building' (which line 601 introduces without saying whether or not that is the meaning of 'fixed').


One of the other options is of course to remove the word 'Fixed', which likely solves the problem as well.

graphics/rcd/lang/en_GB.txt
```diff
@@ -185,6 +196,8 @@ strings {
 
 		// Ride build window strings.
 		RIDE_BUILD_TITLEBAR_SHOP:         "Place shop";
+		RIDE_BUILD_TITLEBAR_GENTLE:       "Place gentle ride";
+		RIDE_BUILD_TITLEBAR_THRILL:       "Place thrill ride";
```
Alberth289346, line 200: Apparently, we have both gentle and thrill rides. The FGTR block doesn't say if the spooky manor is gentle or thrill. Should it?

graphics/rcd/lang/en_GB.txt
```diff
@@ -244,4 +257,14 @@ strings {
 		NAME_TYPE:        "Basic rollercoaster";
 		DESCRIPTION_TYPE: "Basic steel rollercoaster ride";
 	}
+
+	stringtexts("spooky-manor") {
+		NAME_INSTANCE1:        "Palace of Fear";
+		NAME_INSTANCE2:        "Aisles of Darkness";
+		NAME_TYPE:             "Spooky Manor";
+		DESCRIPTION_TYPE:      "A dark, creepy mansion that terrifies any guests brave enough to enter it.";
```
Alberth289346: nice :)

Typo: "any guest[s]"

"Any" implies singular form.

src/ride_type.h
```diff
@@ -75,7 +76,7 @@ class RideType {
 	virtual RideInstance *CreateInstance() const = 0;
 	void SetupStrings(TextData *text, StringID base, StringID start, StringID end, StringID name, StringID desc);
 
-	const RideTypeKind kind; ///< Kind of ride type.
+	RideTypeKind kind;       ///< Kind of ride type.
```
Alberth289346, line 79: You can make it `const` if you initialize it before entering the constructor, ie

`RideType(RideTypeKind rtk) : kind(rtk) { ... }`

src/ride_build.cpp
```diff
@@ -168,67 +175,96 @@ void RideBuildWindow::OnClick(WidgetNumber wid_num, const Point16 &pos)
...
+ * @return The space is suited to build the ride.
  */
 static bool CheckSufficientVerticalSpace(const XYZPoint16& position, const int8 height)
 {
 	for (int8 h = 0; h <= height; ++h) {
 		const Voxel *v = _world.GetVoxel(position + XYZPoint16(0, 0, h));
-		if (v != nullptr && !v->CanPlaceInstance()) return false;
+		if (v == nullptr) continue;
+		if (!v->CanPlaceInstance() || v->GetGroundSlope() != SL_FLAT) return false;
+		if (h > 0 && v->GetGroundType() != GTP_INVALID) return false;
```
Alberth289346, line 189: Bit weird to first query the ground slope and if flat, only then ask whether there is any ground.

You may be lucky that non-existing ground is always flat, but I wouldn't rely on it. Please first check if there is valid ground before you discuss slopes of it.

An empty line after line 187 makes more clear that there are 2 different kinds of checks here. (Give readers visible clues that you change the topic, so even if one is  skimming the code it is less likely to miss the difference between "continue" and "return"

src/ride_build.cpp
```diff
@@ -168,67 +175,96 @@ void RideBuildWindow::OnClick(WidgetNumber wid_num, const Point16 &pos)
...
- * Checks whether the air space above the ground at the given location is suited to place a shop of the given height.
+ * Checks whether the air space above the ground at the given location is suited to place a fixed ride of the given height.
  * @param position Coordinate of the base voxel.
  * @param height Height of the object.
- * @return The space is suited to build the shop.
+ * @return The space is suited to build the ride.
  */
 static bool CheckSufficientVerticalSpace(const XYZPoint16& position, const int8 height)
 {
 	for (int8 h = 0; h <= height; ++h) {
```
Alberth289346: Reading this again, It looks wrong now.

Values `height`, `width`, and `length` are always exclusive when counting from `0` (`height==1` means 1 voxel only, that is, only voxel number `0`). If you meant the inclusive upper bound here, use `last` or `highest`.

I see you're using pre-incrementing. I always use post-incrementing as I find that more natural to read.
Likely in the past it had performance consequences, but nowadays, rewriting this to the best form is a job that the compiler does better than us.
(To be clear, I am **not** asking for a change in incrementing style.)

src/rcdgen/nodes.cpp
```diff
@@ -974,6 +974,56 @@ int SHOPBlock::Write(FileWriter *fw)
...
+{
+	FileBlock *fb = new FileBlock;
+	fb->StartSave(this->blk_name, this->version, 61 + 17 * (this->ride_width_x * this->ride_width_y) - 12);
+	fb->SaveUInt8(this->is_thrill_ride ? 1 : 0);
+	fb->SaveUInt16(this->tile_width);
+	fb->SaveUInt8(this->ride_width_x);
+	fb->SaveUInt8(this->ride_width_y);
+	for (int8 x = 0; x < this->ride_width_x; ++x) {
+		for (int8 y = 0; y < this->ride_width_y; ++y) {
+			fb->SaveUInt8(this->heights[x * ride_width_y + y]);
```
Alberth289346: The RCD block description doesn't say what the major and minor indices are, which is useful if you have a ride of at least 2x2.

src/rcdgen/check_data.cpp
```diff
@@ -1763,6 +1763,69 @@ static std::shared_ptr<SHOPBlock> ConvertSHOPNode(std::shared_ptr<NodeGroup> ng)
...
+		return nullptr;
+	}
+	block->tile_width = vals.GetNumber("tile_width");
+	block->ride_width_x = vals.GetNumber("ride_width_x");
+	block->ride_width_y = vals.GetNumber("ride_width_y");
+	block->heights.reset(new int8[block->ride_width_x * block->ride_width_y]);
+	block->ne_views.reset(new std::shared_ptr<SpriteBlock>[block->ride_width_x * block->ride_width_y]);
+	block->se_views.reset(new std::shared_ptr<SpriteBlock>[block->ride_width_x * block->ride_width_y]);
+	block->nw_views.reset(new std::shared_ptr<SpriteBlock>[block->ride_width_x * block->ride_width_y]);
+	block->sw_views.reset(new std::shared_ptr<SpriteBlock>[block->ride_width_x * block->ride_width_y]);
```
Alberth289346: It doesn't make a difference in performance, but it's useful if you don't copy/paste computations. Make a new variable the contains the size, and use that instead. It makes it easier to see that all arrays are equally large. (Currently, you have to carefully check each line to be exactly the same as the previous line and we are very bad at that, so avoid forcing people into that.)

One of the options for `views` is to make an array of views.
Not sure it's worth the hassle though, as now it's perfectly clear which variable contains which views, something that you loose with `block->views[i]`.

src/rcdgen/check_data.cpp
```diff
@@ -1763,6 +1763,69 @@ static std::shared_ptr<SHOPBlock> ConvertSHOPNode(std::shared_ptr<NodeGroup> ng)
...
+	ExpandNoExpression(ng->exprs, ng->pos, "FGTR");
+	auto block = std::make_shared<FGTRBlock>();
+
+	Values vals("FGTR", ng->pos);
+	vals.PrepareNamedValues(ng->values, true, true);
+
+	const std::string category = vals.GetString("category");
+	if (category == "gentle") block->is_thrill_ride = false;
+	else if (category == "thrill") block->is_thrill_ride = true;
+	else {
```
Alberth289346: if/else always has curly braces.

Only single line statements don't have curly braces.

graphics/rcd/lang/en_GB.txt
Alberth289346: Found it already, I missed it. Sorry.

src/ride_type.h
Noordfrees: The information whether we have a gentle or thrill ride only becomes available during `Load()`, so we still need the possibility to change the `kind` after the ctor.

doc/data_format.rst
```diff
@@ -596,40 +596,40 @@ Version history
...
+  33+s+16*s*(a+1)   4          1-       nw preview.
+  37+s+16*s*(a+1)   4          1-       First recolouring specification.
+  41+s+16*s*(a+1)   4          1-       Second recolouring specification.
+  45+s+16*s*(a+1)   4          1-       Third recolouring specification.
+  49+s+16*s*(a+1)   4          1-       Entrance fee.
+  53+s+16*s*(a+1)   4          1-       Monthly cost of having the ride.
+  57+s+16*s*(a+1)   4          1-       Additional monthly cost of having an opened ride.
+  61+s+16*s*(a+1)   4          1-       Text of the ride (reference to a TEXT block).
+  65+s+16*s*(a+1)                       Total length.
+=================  ==========  =======  ===================================================================================
```
Alberth289346: There are several things I'd like to discuss here, from simple to quite deep. It's not so much about code itself.

### Phases
You introduced 'phase' as new concept, and interestingly named it `a`. Technically nothing wrong with it, but imagine you didn't write this, and instead you got this definition from someone that wants you to code a reader or writer program for it. You haven't seen it before, and you're trying to understand what all the stuff means.
Now, in that situation, would you prefer `a` or `p`? (formulate an answer for yourself)

The letter `p` is much more obviously connected to 'phase' in my view. In other words, by picking an 'good' letter, you make it simpler for the reader to understand what you wrote.

Now for one letter, the difference isn't very big, but imagine what happens if you have 20 variables and concepts, and the variable names and concepts are not easy to connect to each other. Any chance of understanding such an explanation you think?

Code and text are very not very different, both describe things that the author wants to explain to the readers, both are easier to grasp when the author aims for maximum readability.

### Phases + 1
You defined phases, and then ended up with `a+1` all over the place. You can find it in this block of text, but also in the code. In hind sight, do you think the phases definition was a good one? Are there better definitions that avoid the  `+1`?

### Frames
The 'phases' definition is not usual in animations. The normal notion being used is 'frames', where a frame is a single 'picture' (a set of images displaying a point in time in the animation). The reason is that graphics artists produces 'pictures', so they are interested in how many there are.
A non-animated image is 1 frame, with 2 frames you alternate between 2 images, etc.

A nice side-effect when you switch to frames instead of phases is that your `a+1` disappears, since frame count is equal to the number of full-building images that you have.

### Purpose of these block definitions
I am not sure you fully understand why we have these block definitions, so let's discuss that.

These definitions fully document how to interpret each and every bit in an RCD file. The text files are the one and only authoritative definition. If our code doesn't do what this text says, the code is wrong.

In this way you get a clean separation between code that generates RCD files, and code that reads and displays their contents. Someone can take these text files, and code a new `rcdgen` if they want. At the reading side, someone can take existing RCD files, and code a new RCT program from these text files, or he can load the RCD files into OpenRCT2, or whatever other program. The RCD file format is a documented open format, available to all. It's not an internal FreeRCT-only file format.

In order to make it work, the definition in these files must be complete. Any question about meaning of any bit in any RCD file must be answered by these text files.

This is why I said there has to be documentation of how the major and minor indices work in the arrays of images in this block. I said that somewhere in the code because that's where I realized the information was missing. You documented it in the code and not here, so the current description is not yet complete. I cannot read in this document what byte `21+s+15` is used for, if `x==2`, `y==2`, `a==3`.

(Next I'll be throwing lots of new questions to you. While I think they deserve consideration, and eventually a satisfactory answer, their main purpose now is an introduction to the next section.)

Another thing that seems missing is the purpose of these arrays. Take for example array `21+s`. It's an animation. So when is the animation displayed? Always? When the ride is running only? Is there also an animation to play while loading new passengers?

Does the current solution cover moving rides, eg the ship swaying back and forth? Starting slowly, then picking up speed, swinging back and forth for a while, and slowing down again? Is this the only movement pattern for all rides, or are there other patterns as well? Is behavior pattern intended to be covered in the FTGR block?


### Reflection and tomorrow needs
A while ago I asked a question in the discussion issue, and you almost immediately jumped to the solution "FixedRide class". In this merge request you created a new block with one static view first and coded it. I asked about animations for moving rides, and you added a number of phases, and coded it.

You seem very keen to think and code solutions. It looks like you pick the first solution that pops up in your head, and code it. Every time I pose a problem to you, you quickly pick a solution and code it.

As I tried demonstrating by the bunch of next questions above, jumping on the first quick answer and code it is not always the fastest way to move forward. You coded two solutions you may have avoided spending time on by more thinking before grabbing the keyboard.

I found out a long time ago that the first solution that pops up in your head is almost always not the best solution. There are always things you don't take into account in the first solution. It pays to think a bit longer on the problem. Does the solution make sense? Does it solve everything I want it to cover? Are there any points where I can improve the solution? Are there things out there where my solution fails? If so, is that acceptable or should the solution be enhanced?

In other words, once you have a solution, review it for sanity and major holes that will come back and hunt you tomorrow. By doing that, you'll often find better solutions. Right there you're saving time by not coding the inferior solution first. It makes you a more effective programmer.

Unlike what people generally think, "programming" is the game of properly understanding the problem and its solutions in the complete context, it's not about how fast you can code.

I realize this is a big one, it will take a life time to develop it for yourself.

### Next
So, what to do with this merge request then eh?

Well, the minimum that is required is to make the current definition of the FGTR complete. I also think switching to frames is beneficial. With that, a lot of follow up questions become out of scope for this patch (which is fine to me).

If you want to aim for more, we should probably first discuss the scope (what rides are supposed to be covered by it), and decide on the block definitions. Once that is established, code can be written.

Maybe the block should be split into several parts, so it doesn't become so large.

Noordfrees: Allow me first to answer the more in-depth questions. I make it a rule to not propose anything – neither in discussion nor in code – unless I already thought about how to implement it. First, I have an idea "I want to code thrill rides". Then I think *how* they might work internally and develop all related concepts. There's a big distinction between 'tracked' and 'fixed' rides – I decided to work only on the 'fixed' type now, and the 'tracked' type (e.g. maze, gokarts) is an entirely different project altogether. Next, what does such a ride need? Animations, alternating idle/working phases, breakdowns&reliability – my concept already has all of this. (There is no concept yet for actually letting persons interact with rides, but as they don't interact with coasters either and there are no entrances and exits anywhere yet, I believe this is not an immediate necessity.)

But one of the first lessons I learnt in open-source development is that large branches are reviewer-hostile, and tend to get very hard to maintain if there are review comments to address (and there *always* are). So I split my big concept into a series of standalone units. 
First I implement the basic functionality: A ride which doesn't *do* anything yet. 
Then a second branch to distinguish the four main cases of animations: The "idle" animation (when the ride is not doing anything), the "start" animation (which is a series of sprites displayed once when the ride begins to work, e.g. the ship picking up speed), the "working" animation (which loops around during the main part of the ride's working – the ship swinging from side to side), and the "end" animation (like the start animation in reverse, displayed once at the end of the working cycle – the ship coming slowly to a standstill). 
Then a third branch to regulate the timing of the animations and implement a fixed cycle that can actually be seen. And so on. 

At a reviewer's request I can also implement parts of such things in an earlier branch. That's why I had the answers to your questions ready in code immediately. To me it sounds like you'd prefer me to put it all into a single branch. Since FreeRCT is not being heavily developed currently there is little risk of significant merge conflicts appearing, so this would be fine for me. I had assumed that you'd prefer reviewing several small branches over one huge one, but it's your choice if I should keep this PR "small" and complete the implementation in several other PRs, or put everything in here.

---

Splitting the FGTR block is a very good suggestion, I will do this. I'll also change the wording 'phases' → 'frames'. I think a good approach would be a new block type GTRA that defines an animation by specifying the number of frames, the number of voxels in x and y direction, and also FPS; the FGTR block will then contain references to several such blocks. Orientation could be handled by either having 4 GTRA blocks per animation in the FGTR, or by having all 4 directions for an animation in the GTRA block, or by introducing another block type GTAS that groups 4 GTRA blocks. What do you think? 
(There are also ANIM and ANSP blocks but those seem quite person-specific and not really suited to the needs of gentle/thrill rides.) This change will allow getting rid of most variables in the FGTR's bit table. I also understand your point about the careful documentation; writing code to fit a previously defined standard is a way of coding that I am not yet entirely familiar with and I hope I'm not straining your patience too much.

Finally, I am curious which code paths exactly you mean by 

> You coded two solutions you may have avoided spending time on by more thinking before grabbing the keyboard.

To me, every time I addressed your review it was (in my eyes) either codestyle, plain errors that needed fixing, a logical optimization that did not change the intention behind the code, or the early implementation of what I meant to add or extend anyway in another branch.

src/fixed_ride_type.cpp
```diff
@@ -0,0 +1,233 @@
...
+	/* Images and texts are handled by the sprite collector, no need to release its memory here. */
+}
+
+const ImageData *FixedRideType::GetView(uint8 orientation) const
+{
+	return (orientation < 4) ? this->previews[orientation] : nullptr;
+}
+
+/**
+ * \fn int GetRideCapacity() const
```
Alberth289346:
```suggestion
 * \fn int FixedRideType::GetRideCapacity() const
```

doc/data_format.rst
Alberth289346: > I am curious which code paths exactly you mean by [not code intermediate solutions]

It seems to me that coding eg a stationary display first, only to remove it and replace it with something else is a waste of time. However, your reasoning is correct, immediately jumping to the final implementation makes life more complicated too.
So this might just be down to personal preferences, and different backgrounds.
You have an interesting view on the matter, I must try that too, thanks!

There might also be the issue of not knowing what you know. I saw you adding "a" animation phases, and thought: "now wait a minute, this isn't enough, where is &lt;various things you need&gt;?". You had it planned as future extension, but I didn't know that. That lack of knowledge made it hard to understand the aim of the current branch, and how it fits into the future.

As for next branches, I am fine with the current approach, no need to change it.

----

So what remains in this branch I think is adding something like

    A view consists of an idle frame, followed by 0 or more running frames. A single frame consists
    of a tile image reference to each tile of the ride, with x as the minor index and y as the major index.

to the documentation.
Not sure if `image reference` is the good word.

That looks like it defines everything, I think. Do you agree?

----

> there are no entrances and exits anywhere yet

The caroussel branch graphics in my repository does include entry and exit houses. But otherwise, there is nothing yet indeed.

----

I agree that existing animations are indeed very person-specific and not useful here.

> I think a good approach would be a new block type GTRA that defines an animation by specifying the number of frames, the number of voxels in x and y direction, and also FPS; the FGTR block will then contain references to several such blocks. Orientation could be handled by either having 4 GTRA blocks per animation in the FGTR, or by having all 4 directions for an animation in the GTRA block, or by introducing another block type GTAS that groups 4 GTRA blocks. What do you think?

There are 2 parts I think. One is "dynamic" behavior" (ie the timing), and one is pixel data (images). The common point between both is a frame of the entire ride in all 4 orientations. Let's call this F here.

Towards behavior, I think a good building block is SF, which is a sequence of F without timing (an untimed frame set), and then another block TIM for timing sequences of several SFs. For simplifying ride specs, for each SF, TIM should likely have a repetition count, an inter-frame time, and a way to optionally increase or decrease frame speed for each cycle of an SF. Finally, we likely want to control ride length in the user interface, so some repetition counts should be derivable from ride length.

Towards pixel data, it's useless to have individual F blocks know about voxel counts, since they are all the same. This also holds for SF blocks, so the first useful spot for voxel counts is in TIM. I don't remember how I coded a voxel image, but one thing that worries me is size of the entire pixel data set. It would be really nice if you could re-use the same voxel data at several places in one or more Fs. Eg the caroussel has 1 set of voxel images (base colour and recolour masks), used for all 4 views.

It may be useful not to glue this too tight to gentle/thrill rides. Basically TIM describes an (any) animation and there is very little ride-specific information in it, it might be useful at other spots too with small changes, eg in scenery.

EDIT: While F is useful as notion here, and perhaps even good for specifying an animation, it makes the SF block having little added value, I feel it may be beneficial to shift data somewhere.
I see two options, drop F and merge the Fs into SF, or drop SF and have a sequence of sequences timed Fs in TIM.

----

> I also understand your point about the careful documentation; writing code to fit a previously defined standard is a way of coding that I am not yet entirely familiar with and I hope I'm not straining your patience too much.

Don't know how much TDD you're doing, but to a large part it shares the same ideas. In TDD you write a (failing) test first for some new feature, then implement code to make the test give an OK result. Here, you write the documentation first, then implement it using only information that's in the document. You cannot execute the document for an OK result, so it comes down to you realizing you need information in the implementation that's not in the document.

And you aren't near the end of my patience at all :)

graphics/rcd/gentle_thrill_rides.txt
```diff
@@ -0,0 +1,52 @@
...
+		height_1_3: 40;
+		height_2_0: 3;
+		height_2_1: 2;
+		height_2_2: 7;
+		height_2_3: 30;
+
+		(se, sw, nw, ne): spritefiles {
+			x_base: 1; y_base: 0;
+			file: "../sprites/ride_shops/toilet/64_000{seq(1..4,1)}.png";
+			recolour: "../sprites/ride_shops/toilet/64p_000{seq(1..4,1)}.png";
```
Alberth289346: toilet?

Noordfrees: This is the view of the file from an earlier commit, before I added the spritesheets…

PR #163, opened 2021-02-02
Low German (nds) Translation

Noordfrees:


doc/data_format.rst
```diff
@@ -306,6 +306,7 @@ en_US  United States.
 es_ES  Spain.
 nl_NL  The Netherlands.
 sv_SE  Sweden.
+nds    Low German.
```
Alberth289346: Since it's part of Germany, it would be `de_NDS`, and to not make a mess and make it difficult to find names of translations, it should be alphabetically ordered, ie below line 303.

Alberth289346: Perhaps `name of language` should be `country` or so to avoid confusion?

Noordfrees: The official language of Germany is German (`de_DE`). Low German (`nds`) is a minority language spoken in the northern half of Germany and also in parts of the Netherlands and Denmark. Sometimes it's also called `nds_DE`, but this is uncommon; usually it's called `nds` without a country suffix. Low German is not the official language of any country.
https://en.wikipedia.org/wiki/Low_German

Alberth289346: The system that we had so far was `country_language`. To me, the country prefix is just a global geographic location. Whether or not that language is official doesn't matter, as far as I am concerned.

The purpose of the country prefix is to make finding back a given language simpler. For 5 languages the problem isn't real, but things grow, and it should stay simple even if we end up with 50 or 200 languages, and the reader doesn't know what the 3 letter name(s) mean. It may not be 100%, but `de_` is loads more precise where it is used than "anywhere in the world".

You cannot just ignore the rules stated in documentation and add a new kind of language name. Text clearly says `country_language` as form of a translation name so that's what it currently is. It's the same as code-style, you cannot just ignore the defined code style of the project and use your own style. Things have to stay consistent in the text as well as in the code.

So there are 2 options here.
- First option is you make your change consistent with the described current system.
- Second option is to modify the rules, and adapt all text that states or uses the rules so it's consistent with itself again. Also some code or files have to be changed as well, as it uses the names defined in the text. (I don't think it uses the country codes on its own, but not sure.)

There are several ways to change it, we could switch to some new form of location information separate from the language name, or use a different naming scheme of languages, or something else.

The country codes are from an international standard that maintains the official set of names (We don't want to explain that `de` means Germany etc, nor do we want to define what a proper name for a country is.)
I looked for languages attached to countries some days ago as a standard, but couldn't find a good list.

Today I had another go based on another way to find such a list, and found ISO 639-3 at https://iso639-3.sil.org/ 
Your `nds` language in it, but the naming system is different, so if we switch to it, all our other language names are wrong.
It doesn't have geographic information unfortunately, but as it is an official list, it should describe a language in a findable way (didn't look much at it, severe lack of time currently).

I am open to other suggestions, but it should be scalable to a few hundred languages without effort, eliminate arguments about proper names for languages, and include some way to make the language findable even if you don't know the few hundred names.

Noordfrees: > `country_language`

Now I'm confused. The two English languages are called "en_GB" and "en_US" so isn't it `language_country` rather than `country_language`?

Alberth289346: Fair enough :)

Alberth289346:
Thanks.

PR #164, opened 2021-02-02
Gentle/Thrill Rides: Idle/Working Phases

Noordfrees:
Gentle/thrill rides now alternate between an "idle" and a "working" phase. The durations of each phase are set in the RCD file. The working phase is fully animated.
Rides also have a capacity now, which is set in the RCD file. Guests still can't actually enter these rides yet (which I intend to change soon though).

Added a new thrill ride called the Ship Swing. The glitch in its working animation is a bug in my Blender model (which became rather too huge), not in the branch.

I added a "sheet" feature to the TIMA block generator to generate the FSET blocks automatically from a spritesheet, rather than having to define each FSET block by hand, and I also added a utility for easier generation of such a spritesheet from raw blender renders.


doc/data_format.rst
```diff
@@ -678,14 +678,18 @@ Offset     Length  Version  Description
   59+s      4       1-      Entrance fee.
   63+s      4       1-      Monthly cost of having the ride.
   67+s      4       1-      Additional monthly cost of having an opened ride.
-  71+s      4       1-      Text of the ride (reference to a TEXT block).
-  75+s                      Total length.
+  71+s      4       2-      Number of guests that can use the ride at the same time.
+  75+s      4       2-      Duration of the ride's idle phase in milliseconds.
+  79+s      4       2-      Total duration of the ride's working phase in milliseconds.
```
Alberth289346: I am guessing that idle duration and working duration are assumed to be fixed values for now, and they will change in the future.

However, you also have animation timing here, so there needs to be statement what happens if the animation specification and the duration here are in conflict.

src/fixed_ride_type.h
```diff
@@ -21,7 +21,12 @@ class FixedRideType : public RideType {
 	explicit FixedRideType(RideTypeKind kind);
 	~FixedRideType();
 
-	virtual int GetRideCapacity() const = 0;
+	/* Information about how many guests can use the ride at the same time. */
+	struct RideCapacity {
+		int number_of_batches; ///< How many batches of guests fit into the ride.
+		int guests_per_batch;  ///< How many guests may be in each batch.
```
Alberth289346, line 27: Shouldn't there be 2 values for "number of guests in the ride" in the FGTR block?

Please keep the block definitions and code in sync.

src/video.cpp
```diff
@@ -624,21 +624,26 @@ static void Blit32bppImages(const ClippedRectangle &cr, int32 x_base, int32 y_ba
...
+			const bool is_in_drawing_region = (src_base >= cr.address && src_base < cr.address + cr.width * cr.height);
 			switch (mode >> 6) {
 				case 0: // Fully opaque pixels.
 					mode &= 0x3F;
 					if (shift == GS_SEMI_TRANSPARENT) {
 						src += 3 * mode;
 						for (; mode > 0; mode--) {
-							uint32 ndest = BlendPixels(255, 255, 255, *src_base, OPACITY_SEMI_TRANSPARENT);
-							BlitPixel(cr, src_base, xpos, ypos, numx, numy, spr->width, spr->height, ndest);
+							if (is_in_drawing_region) {
```
Alberth289346: This condition doesn't change in the loop so it shouldn't be here, also, it doesn't do full protection as `src_base` changes here.

src/fixed_ride_type.h
Noordfrees: A fixed ride can only ever contain 1 batch of people, so there's no need to put this in the RCD imho. The only two exceptions I can think of in the original RCT are the ferris wheel, which is such a *special* ride in every way that it will likely need its own class when (if?) it gets implemented, and Tracked Rides (maze, gokarts, etc), which are technically a coaster derivate and also not covered by FixedRideType.

doc/data_format.rst
Noordfrees: Added info about the requirements. There should be no need to change the specification here in the future though:
- The duration how long the ride idles can be heavily controlled by the player in RCT. You can set a minimum and maximum idle time, and specify that the ride shall start working on full load or half load etc. Any such setting overrides the duration from the RCD. But if the player says Any Load with no min/max time then the program should use the default value stated in the RCD.
- For the working phase, some rides allow the user to specify how many cycles the ride takes. Then the working duration from the RCD refers to 1 cycle, and the actual duration is the product of this base duration and the player's setting. I could already add space for the min and max numbers of cycles per working phase to the file definition, but as this feature is not of very high priority to me they would lie around unused for a while so I'd rather wait with adding them until several more features have been implemented. (The next steps in my plans are to allow building entrances and exits, and then to let guests actually enter rides.)

src/fixed_ride_type.h
Alberth289346: I usually don't stop where the original RCT ran out of time or money.

The two examples I had in mind for this were all coaster rides, and a movie theater where several movies can be shown at the same time. Your haunted house may also be an example. I can imagine several small groups of people visit the ride at the same time, one group after the other.

Not sure why you consider the Ferris Wheel special? Just make an animation to turn the wheel to the next cabin (which could be eg 11/6 further down the wheel to fake the idea of turning the wheel between changing passengers). As long as all cabins get a turn to exchange passengers, it's fine. (Where 11/6 is just  an example, didn't verify all cabins actually get a turn then.)

Alberth289346: Actually, for `N` cabins, the simplest is to turn `N-1` cabins or `N+1` cabins, which trivially gives all cabins a turn.

Noordfrees: Okay, I've added this to the RCD, and also split up the spooky house's group. However I had to add the requirement that rides with more than 1 batch mustn't have a real working animation, as then the ride's idle/working phases (and the point in time in the working animation) are no longer well-defined.

For the ferris wheel, this implementation would be possible though it would have the weird side-effect that each guest would first stand on the ground waiting for a cabin, then ride one round, then stand around on the ground waiting for the remaining guests to exit. I would expect that he'd spend the first waiting time in the queue (not enter the ride until a free cabin is on the ground) and exit the ride immediately when exiting the cabin. But I guess that this could be considered an eye-candy detail so ok…

Alberth289346: Current decisions are good enough for now, I think.

I don't see "idle" and "empty" as the same things, although it holds for a large number of rides. Also "working" and loading/unloading are perhaps less mutual exclusive than currently assumed, eg toilets with finite space inside.
But it's all for the future.

PR #165, opened 2021-02-04
Automatically detect system language

Noordfrees:
Try to automatically detect the system language on startup by parsing the environment variables `LANG` and `LANGUAGE`. If a user specified multiple languages in `LANGUAGE`, pick the first for which we have translations. Falls back to `en_GB` if no suited language can be detected.


Alberth289346:
Having a non-system variable that takes precedence is likely useful, eg `FREERCT_LANG` or so. Not everybody uses one language for everything, and having to switch your system language so you can check a bug fix in Chineese rendering doesn't sound like a terribly good plan to me :)

The better option is to add the used language to the configuration file. Environment variables are just horrible to manage and quite system-specific (that is, non-portable).

Had a quick look at the code, not sure why you add such a long function as a lambda.

src/freerct.cpp
```diff
@@ -125,6 +125,16 @@ int freerct_main(int argc, char **argv)
 		                "medium-path = /usr/share/fonts/gnu-free/FreeSans.ttf\n");
 		return 1;
 	}
+	/* Overwrite the default language settings if the user specified a custom language in the config file. */
+	const char *preferred_language = cfg_file.GetValue("language", "language");
+	if (preferred_language != nullptr) {
+		int index = GetLanguageIndex(preferred_language);
+		if (index < 0) {
+			fprintf(stderr, "Warning: Language '%s' is not known, ignoring\n");
```
Alberth289346: Missing a parameter for the `%s`.

----

In addition, in its current form, the message is useless information. It currently basically just says "it doesn't work", which is really old news for a user that tries to set or change a language.

Never just say "something fails", always try to provide sufficient information such that a user can find the cause and fix it. Here you would say you read the language name from the configuration file.

Further extensions on this path are stating the file path of that file, listing the currently known languages, try a case-insensitive check, try a fuzzy match to give a hint to the use ("maybe you meant UVW language?"), and more.

It may be more work, but you'll make many users happy, get fewer help requests into your inbox, and eventually even you won't remember where the message is coming from, and you'll be glad for the provided assistance.

src/language.cpp
```diff
@@ -76,6 +76,53 @@ int GetLanguageIndex(const char *lang_name)
...
+		double s = std::max(name.size(), lang_name.size()) - length;
+		s *= -1;
+		for (int j = 0; j < length; ++j) {
+			const char c1 = lang_name.at(j);
+			const char c2 = name.at(j);
+			if (c1 == c2) {
+				s++;
+			} else if (c1 >= 'a' && c1 <= 'z' && c2 >= 'A' && c2 <= 'Z' && c1 + 'A' - 'a' == c2) {
+				s += 0.5;
+			} else if (c2 >= 'a' && c2 <= 'z' && c1 >= 'A' && c1 <= 'Z' && c2 + 'A' - 'a' == c1) {
```
Alberth289346: If one side is a letter, then the final (5th) check at the line already covers that the other side is also a letter (in upper case), or you never get equality. Thus, the 3rd and 4th check aren't needed.

src/freerct.cpp
```diff
@@ -125,14 +136,34 @@ int freerct_main(int argc, char **argv)
...
+	}
+	if (!language_set) {
+		preferred_language = cfg_file.GetValue("language", "language");
+		if (preferred_language != nullptr) {
+			int index = GetLanguageIndex(preferred_language);
+			if (index < 0) {
+				fprintf(stderr, "The language '%s' set in the configuration file (freerct.cfg) is not known.\n", preferred_language);
+				const char *similar = GetSimilarLanguage(preferred_language);
+				if (similar != nullptr) fprintf(stderr, "Did you perhaps mean '%s'?\n", similar);
+				fprintf(stderr, "Call 'freerct --help' for a list of all supported languages.\n");
```
Alberth289346: `Call` seems a bit weird to me, perhaps `Use` or `Type` or `Enter` ?

src/freerct.cpp
```diff
@@ -52,8 +53,14 @@ static void PrintUsage()
...
 	printf("Options:\n");
-	printf("  -h, --help           Display this help text and exit\n");
-	printf("  -l, --load [file]    Load game from specified file\n");
+	printf("  -h, --help           Display this help text and exit.\n");
+	printf("  -l, --load [file]    Load game from specified file.\n");
+	printf("  -a, --language lang  Use the specified language.\n");
+	printf("                       Valid languages are:");
+	for (int i = 0; i < LANGUAGE_COUNT; ++i) {
+		printf(" %s", _lang_names[i]);
+	}
```
Alberth289346, line 69: The list of languages can be separate from the `-a` option. (Just add an empty line, then print the text starting from the left margin.

If you want to be fancy, go to the next line after 70 or so characters :)

src/language.cpp
```diff
@@ -76,6 +76,53 @@ int GetLanguageIndex(const char *lang_name)
...
+ */
+const char *GetSimilarLanguage(const std::string& lang_name)
+{
+	const char *best_match = nullptr;
+	double score = 1.5;  // Arbitrary treshold to suppress random matches.
+	for (int i = 0; i < LANGUAGE_COUNT; ++i) {
+		const std::string name(_lang_names[i]);
+		const int length = std::min(name.size(), lang_name.size());
+		double s = std::max(name.size(), lang_name.size()) - length;
+		s *= -1;
```
Alberth289346: Not sure why you don't wrap that into the previous line, ie
``` cpp
double s = -(std::max(name.size(), lang_name.size()) - length);
```

src/language.cpp
```diff
@@ -76,6 +76,53 @@ int GetLanguageIndex(const char *lang_name)
...
+ * @param lang_name Name to compare to.
+ * @return Most similar language name, or \c nullptr if no language has a similar name.
+ */
+const char *GetSimilarLanguage(const std::string& lang_name)
+{
+	const char *best_match = nullptr;
+	double score = 1.5;  // Arbitrary treshold to suppress random matches.
+	for (int i = 0; i < LANGUAGE_COUNT; ++i) {
+		const std::string name(_lang_names[i]);
+		const int length = std::min(name.size(), lang_name.size());
```
Alberth289346: `common_length` looks like the better name for this.

PR #166, opened 2021-02-07
Ride entrances and exits

Noordfrees:
Gentle/thrill rides now have an entrance and an exit. These can be placed by the user from the ride management window. 

There are two entrance styles and two exit styles available, which can be selected for each ride individually in the management window. The "Generic" style is taken from https://github.com/Alberth289346/FreeRCT/tree/new-graphics/graphics/sprites/ride_bits; the "Alternative" style is a GIMPed and recolourable version of the same graphics (and should be considered only a proof-of-concept placeholder for the recolouring feature).

Guests can enter a G/T ride through the entrance, and leave through the exit when the guest batch is done. If the ride can not allow a guest in already, he will wait patiently outside until the ride admits him. A queue path leading up to the entrance is optional, it will work with or without one.

Next TODOs I'm going to address in the next branches:
- Let guests wait in a nice, orderly queue line, rather than all on exactly the same spot as they do now.
- Generalize the entrance/exit-related code to also allow placing coaster entrances and exits.


Alberth289346:
Do you understand the entry and exits have 2 layers of graphics?  images 1..4 are the front layer, and 5..8 are the background layer, both for 4 directions (1 matches with 5, 2 with 6, etc).

What you do is first paint the background, then paint all people there, and finally paint the foreground, so only people that should be (possibly partly) visible can be seen. Afaik, this is already implemented. I don't remember how exactly, but I would guess you can specify a desired layer at a tile when passing the sprite to the sprite sorter.

PR #167, opened 2021-02-11
Implement queuing behaviour for guests

Noordfrees:
Let guests line up in a neat queue line when waiting for a ride. A queue path can hold approximately 4 guests per tile. Guests will refuse to enter a queue if the queue line is completely filled. For rides where the entrance is directly at a normal path, at most 1 guest can queue.

grafik

### DIFF
--- a/src/person.cpp
+++ b/src/person.cpp
@@ -23,6 +23,9 @@
 
 static PersonTypeData _person_type_datas[PERSON_TYPE_COUNT]; ///< Data about each type of person.
 
+static const int QUEUE_DISTANCE = 64;  // The pixel distance between two guests queuing for a ride.
+assert_compile(256 % QUEUE_DISTANCE == 0);
+
 /**
  * Construct a recolour mapping of this person type.
  * @return The constructed recolouring.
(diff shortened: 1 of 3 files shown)


src/person.cpp
```diff
@@ -895,6 +911,63 @@ void Person::DeActivate(AnimateResult ar)
...
+			XYZPoint16(vox_pos.x, vox_pos.y + 1, vox_pos.z),
+			XYZPoint16(vox_pos.x, vox_pos.y - 1, vox_pos.z)}) {
+		if (!IsVoxelstackInsideWorld(vx.x, vx.y)) continue;
+
+		for (const XYZPoint16& checkme : {vx, XYZPoint16(vx.x, vx.y, vx.z + 1), XYZPoint16(vx.x, vx.y, vx.z - 1)}) {
+			const Voxel *voxel = _world.GetVoxel(checkme);
+			if (voxel == nullptr) continue;
+
+			for (VoxelObject *v = voxel->voxel_objects; v != nullptr; v = v->next_object) {
+				if (v == this) continue;
```
Alberth289346, line 952: This looks somewhat simplistic.

Queuepaths should have only 2 edges to walk on/off the tile, so you know the relevant edges to check of the path tile. Also, it has a slope, so there is no need to brute-force try all levels as far as I can see.
Unless I have been extremely stupid, queue paths use the same system as normal paths, so there should be code to get the neighbour tile.

If the guest isn't near an edge, you can even avoid checking the neighbour tiles completely.

I am also somewhat worried about having several independent queues next to each other, but I guess that works due to "not near enough" for the neighbour check.

From an efficiency point of view, it's likely better to run this check by following the queue paths instead of doing this for each person, don't know how feasible this is.


For an initial step the current code seems sufficient though. Great work.

Issue #72, opened 2015-03-21
Switch person graphics


ghost:
Switch person graphics to the updated earth-bound ones (sorry pillar planet).

PR #171, opened 2021-02-21
Implement entrances and exits for coasters

Noordfrees:
Coasters have entrances and exits now. As they can have multiple stations, every station needs to have exactly one entrance and and one exit, otherwise the ride can't be opened.
The ride's opened/testing/closed state is now fully under the control of the radiobuttons rather than automatically putting the coaster in test mode upon opening its window.
Guests cannot actually visit coasters yet.

### DIFF
--- a/src/coaster.cpp
+++ b/src/coaster.cpp
@@ -654,7 +654,7 @@ void CoasterInstance::CloseRide()
 	RideInstance::CloseRide();
 }
 
-void CoasterInstance::GetSprites(const XYZPoint16 &vox, uint16 voxel_number, uint8 orient, const ImageData *sprites[4]) const
+void CoasterInstance::GetSprites(const XYZPoint16 &vox, uint16 voxel_number, uint8 orient, const ImageData *sprites[4], uint8 *platform) const
 {
 	const CoasterType *ct = this->GetCoasterType();
 
@@ -667,12 +667,14 @@ void CoasterInstance::GetSprites(const XYZPoint16 &vox, uint16 voxel_number, uin
 		const ImageData *const *array = _rides_manager.entrances[this->entrance_type]->images[orientation_index(EntranceExitRotation(vox, nullptr))];
 		sprites[1] = array[0];
 		sprites[2] = array[1];
+		if (platform != nullptr) *platform = PATH_NE_NW_SE_SW;
 		return;
 	}
 	if (this->IsExitLocation(vox)) {
 		const ImageData *const *array = _rides_manager.exits[this->exit_type]->images[orientation_index(EntranceExitRotation(vox, nullptr))];
 		sprites[1] = array[0];
 		sprites[2] = array[1];
+		if (platform != nullptr) *platform = PATH_NE_NW_SE_SW;
 		return;
 	}
 	if (voxel_number == ENTRANCE_OR_EXIT) {
--- a/src/coaster.cpp
+++ b/src/coaster.cpp
@@ -23,6 +23,8 @@ CoasterPlatform _coaster_platforms[CPT_COUNT]; ///< Sprites for the coaster plat
 static CarType _car_types[16]; ///< Car types available to the program (arbitrary limit).
 static uint _used_types = 0;   ///< First free car type in #_car_types.
 
+static const uint16 ENTRANCE_OR_EXIT = INT16_MAX;  ///< Indicates that a voxel belongs to an entrance or exit.
+
 /**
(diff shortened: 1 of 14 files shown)


src/coaster_gui.cpp
```diff
@@ -212,6 +350,101 @@ void CoasterInstanceWindow::SetRadioChecked(WidgetNumber radio, bool checked)
...
+void CoasterInstanceWindow::SelectorMouseMoveEvent(Viewport *vp, const Point16 &pos)
+{
+	const Point32 world_pos = vp->ComputeHorizontalTranslation(vp->rect.width / 2 - pos.x, vp->rect.height / 2 - pos.y);
+	int dx, dy;
+	switch (vp->orientation) {
+		case VOR_NORTH: dx =  1; dy =  1; break;
+		case VOR_WEST:  dx = -1; dy =  1; break;
+		case VOR_SOUTH: dx = -1; dy = -1; break;
+		case VOR_EAST:  dx =  1; dy = -1; break;
+		default: NOT_REACHED();
```
Alberth289346: This looks like it's common code, other occurrences are at `gentle_thrill_ride_gui.cpp` line 365, and `ride_build.cpp` line 291.  Perhaps move it to a common function?

src/coaster_gui.cpp
```diff
@@ -126,21 +165,67 @@ CoasterInstanceWindow::CoasterInstanceWindow(CoasterInstance *ci) : GuiWindow(WC
...
+assert_compile(MAX_RECOLOUR >= 3); ///< Check that the 3 recolourings of a gentle/thrill ride fit in the Recolouring::entries array.
+
+/** Update all recolour buttons of the window. */
+void CoasterInstanceWindow::UpdateRecolourButtons()
+{
+	for (int i = 0; i < 3; i++) {
+		const RecolourEntry &re = this->ci->entrance_recolours.entries[i];
+		this->GetWidget<LeafWidget>(CIW_ENTRANCE_RECOLOUR1 + i)->SetShaded(!re.IsValid());
+	}
+	for (int i = 0; i < 3; i++) {
```
Alberth289346: would a constant for the magic `3` make sense? At least it simplifies finding back the related magic numbers.

src/coaster.cpp
```diff
@@ -23,6 +23,8 @@ CoasterPlatform _coaster_platforms[CPT_COUNT]; ///< Sprites for the coaster plat
 static CarType _car_types[16]; ///< Car types available to the program (arbitrary limit).
 static uint _used_types = 0;   ///< First free car type in #_car_types.
 
+static const uint16 ENTRANCE_OR_EXIT = INT16_MAX;  ///< Indicates that a voxel belongs to an entrance or exit.
```
Alberth289346, line 26: I wonder if it is useful to distinguish between both types by a different value.

src/coaster_gui.cpp
Alberth289346: Perhaps similar to the short arrays in tile_func.cpp at the start, like `_corners_at_edge` (and declaration in `tile.h`). However, being a computation in view orientation, some `orientation_func.cpp` would be better (We only have `orientations.h` currently.)

Alberth289346:
Unfortunately we don't have graphics for cars with people in them.

src/coaster.cpp
Noordfrees: No point IMHO. This special value is used only to recognize that a certain voxel is not occupied by any track piece while drawing the coaster. In all other places we compare whether one of the ride's entrances/exits is located at the coordinate of interest.

Noordfrees:
Yes. I'm *not* looking forward to creating corresponding graphics for the (then downscaled) shipswing :P 
But drawing people on rides is a purely graphical feature so it has pretty low priority IMHO; there are several more important things I want to work on first (guests entering coasters; player-configurable operating settings; breakdowns & handymen; excitement/intensity/nausea ratings). Also #72 should be addressed before starting people-graphics related work (any reason why the images from `graphics/sprites/guests/template_walk8bpp64.png` are not in use yet?).

Alberth289346:
> But drawing people on rides is a purely graphical feature

Not exactly, different people have different colour clothes, so it needs some trickery or additional sprite layers to get different colours for each person in a ride, perhaps some kind of mask image ?

But yeah, tricky :)

> any reason why the images from graphics/sprites/guests/template_walk8bpp64.png are not in use yet?

Nope, it was just waiting until there was time to do it.

PR #172, opened 2021-02-24
Guests on Coasters

Noordfrees:
Coasters can now have multiple trains with multiple cars, configurable from the management window. Trains will wait in the station until the maximum number of guests has entered the train or a timeout (not configurable yet) elapsed. Guests can and do enter and exit coasters now. And if you forgot to test that the trains can actually climb over those hills before opening your ride, your guests will not like the consequences…

### DIFF
--- a/src/window.cpp
+++ b/src/window.cpp
@@ -1143,15 +1143,17 @@ void NotifyChange(WindowTypes wtype, WindowNumber wnumber, ChangeCode code, uint
  * Highlight and raise a window of a given type.
  * @param wtype %Window type to look for.
  * @param wnumber %Window number to look for (use #ALL_WINDOWS_OF_TYPE for any window of type \a wtype).
+ * @param window [out] If the return value is \c true, this will point to the highlighted window.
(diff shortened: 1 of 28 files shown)


src/window.cpp
```diff
@@ -1143,15 +1143,17 @@ void NotifyChange(WindowTypes wtype, WindowNumber wnumber, ChangeCode code, uint
...
  * @ingroup window_group
  */
-bool HighlightWindowByType(WindowTypes wtype, WindowNumber wnumber)
+bool HighlightWindowByType(WindowTypes wtype, WindowNumber wnumber, Window **window)
 {
 	Window *w = GetWindowByType(wtype, wnumber);
 	if (w != nullptr) {
 		_window_manager.RaiseWindow(w);
 		w->SetHighlight(true);
+		if (window != nullptr) *window = w;
```
Alberth289346: Since the return value is a simple boolean based on the found`w`, you could return `w` or `nullptr` and not have the additional `Window **` parameter.

PR #178, opened 2021-03-14
Ride Settings & Breakdowns

Noordfrees:
Allow the user to configure a number of settings for a ride: Entrance fee, maintenance interval, min and max idle time; for gentle rides also the number of working cycles. (The min idle time is not yet considered by fixed rides because I felt this branch is getting too big already; everything else is taken into account.)
Rides break down now occasionally, and call for a mechanic when broken down as well as when the maintenance interval elapsed. Since there are no mechanics yet, rides in need of one are inspected and repaired automagically after a short while.


doc/data_format.rst
```diff
@@ -715,10 +715,16 @@ Offset     Length  Version  Description
...
-  87+s      4       1-      Text of the ride (reference to a TEXT block).
-  91+s                      Total length.
-=========  ======  =======  ===================================================================================
+  83+s      4       2-      Total duration of the ride's working phase per working cycle in milliseconds.
+  87+s      2       3-      Mimimum number of cycles (must be at least 1).
+  89+s      2       3-      Maximum number of cycles (must be at least the minimum number).
+  91+s      2       3-      Default number of cycles (must be at least the minimum and at most the maximum number).
+  93+s      2       3-      Maximum reliability factor (in percents of a percent).
+  95+s      2       3-      Daily reliability decrease (in percents of a percent).
+  97+s      2       3-      Monthly decrease of the maximum reliability (in percents of a percent).
```
Alberth289346: `in percents of a percent`? Interesting notion :)
I'd probably say `range [0..10000]`

I don't understand what "max reliability factor" is. Perhaps add a line of text explaining it.

Alberth289346:
Thanks for adding the explanation, much more clear what values mean now.

PR #181, opened 2021-03-18
Ride rating attributes

Noordfrees:
Rides now have an excitement, intensity and nausea rating. These are not used for anything yet but that will change soon.

For fixed **gentle/thrill** rides, the intensity and nausea score are set in the RCD file. The RCD also defines a base excitement score and two factors which determine how strongly the number of working cycles and the surrounding scenery affect the excitement score. The shipswing's excitement score for instance can be increased by about 200-300 points when placed among rugged hills as compared to placing it on flat ground.

**Coasters** have a more complex approach. They keep statistics on the speed as well as the vertical and horizontal G forces along the track (the current "calculation" of these forces is not exactly related to real physics though), which can also be viewed as graphs in the coaster's GUI. The intensity, nausea and excitement are calculated from these stats. The weighting formulas I use are pretty arbitrary, and chosen to give reasonable results for the various coaster designs I tested. 
Additionally the scenery has a very large influence on the excitement. Even the best coasters have a pretty low score on flat ground, but it can be increased by several hundred points by adding some hills. The proximity of other rides also has a positive effect on coaster excitement.

The logic behind the scoring algorithms is loosely based on
https://rct.fandom.com/wiki/Excitement
https://rct.fandom.com/wiki/Intensity
https://rct.fandom.com/wiki/Nausea

The ratings are divided by 100 when shown in the GUI, which results in a scale comparable to the ratings in the original RCT.

Also, some refactoring and regrouping in the rides management GUIs for all ride types.


doc/data_format.rst
```diff
@@ -722,8 +722,13 @@ Offset     Length  Version  Description
   93+s      2       3-      Maximum reliability (in range 0..10000).
   95+s      2       3-      Daily reliability decrease (in range 0..10000).
   97+s      2       3-      Monthly decrease of the maximum reliability (in range 0..10000).
-  99+s      4       1-      Text of the ride (reference to a TEXT block).
- 103+s                      Total length.
+  99+s      4       4-      Base intensity rating.
+ 103+s      4       4-      Base nausea rating.
+ 107+s      4       4-      Base excitement rating.
+ 111+s      4       4-      Excitement increase per working cycle.
+ 115+s      4       4-      Excitement increase per nearby scenery item.
```
Alberth289346: It would be useful if you add some sort of unit with these fields. Simplest is probably relating values here to the displayed value for the user.

Is "excitement rating" a different unit than "excitement increase"?

In general, if you name some notion "X" (eg at line 725, X = "intensity rating"), then stick to that name everywhere if you intend to refer to the same notion, and use a different name if you intend to refer to a different (but perhaps related) notion. This convention avoids confusion in how many different notions exist, it also makes it more precise in saying which notion is used exactly.

With "increase" at 728/729, are these unsigned values? If not, "change" would be better, as it expresses the option to use a negative value as well. (Although that would be silly here.)

src/coaster.cpp
```diff
@@ -1589,6 +1607,90 @@ bool CoasterInstance::PlaceEntranceOrExit(const XYZPoint16 &pos, const bool entr
...
+	} else {
+		it->second.valid &= valid;
+		it->second.speed        = (it->second.precision * it->second.speed        + speed) / (it->second.precision + 1);
+		it->second.vertical_g   = (it->second.precision * it->second.vertical_g   + vg   ) / (it->second.precision + 1);
+		it->second.horizontal_g = (it->second.precision * it->second.horizontal_g + hg   ) / (it->second.precision + 1);
+		it->second.precision++;
+	}
+}
+
+void CoasterInstance::RecalculateRatings()
```
Alberth289346, line 1633:
```suggestion
/** Recompute ratings of the coaster instance. */
void CoasterInstance::RecalculateRatings()
```

src/coaster.cpp
```diff
@@ -1589,6 +1607,90 @@ bool CoasterInstance::PlaceEntranceOrExit(const XYZPoint16 &pos, const bool entr
...
+{
+	uint64 exc = 100;
+	uint64 iny = 100;
+	uint64 nau = 100;
```
Alberth289346:
```suggestion
	uint64 exc = 100; // Excitement * 100
	uint64 iny = 100; // Intensity * 100
	uint64 nau = 100; // Nausea * 100
```
I hope `* 100` is correct.

src/coaster_gui.cpp
```diff
@@ -143,36 +151,68 @@ static const WidgetPart _coaster_instance_gui_parts[] = {
...
+				Widget(WT_TEXT_PUSHBUTTON, CIW_EDIT, COL_RANGE_DARK_RED), SetData(GUI_COASTER_MANAGER_EDIT, STR_NULL), SetFill(1, 1),
+				Widget(WT_TEXT_PUSHBUTTON, CIW_REMOVE, COL_RANGE_DARK_RED), SetData(GUI_ENTITY_REMOVE, GUI_ENTITY_REMOVE_TOOLTIP),
 	EndContainer(),
 };
 
+/** What kind of data to display in the graph. */
+enum GraphMode {
+	GM_SPEED,
+	GM_VERT_G,
+	GM_HORZ_G,
```
Alberth289346:
```suggestion
	GM_SPEED,  ///< Display speed.
	GM_VERT_G, ///< Display vertical G force.
	GM_HORZ_G, ///< Display horizontal G force.
```

src/geometry.h
```diff
@@ -29,8 +29,32 @@ struct Point {
...
+
 	typedef CT CoordType; ///< Type of the coordinate value.
 
+	/**
+	 * Add a 2D point to this one.
+	 * @param q Second point to add.
+	 * @return Both dimensions added to each other.
+	 */
+	template<typename CT2>
+	inline Point<CT> operator+=(const Point<CT2> &q)
```
Alberth289346: Wouldn't it make sense to return a reference rather than a copy of the value?

I tried to find the rule about this, but failed. I did find examples, but not a discussion.

src/coaster.cpp
Alberth289346: False positive, can be ignored.

Alberth289346:
Really nice work!

src/coaster.cpp
Noordfrees: Documented in `ride_type.cc:372-375`

PR #185, opened 2021-03-28
Add a main menu screen

Noordfrees:
This adds a splashscreen that smoothly fades into the main menu after a few seconds (impatient users can click to skip), and a main menu screen with a background animation and some buttons to start new games, load games, edit settings, and quit. 

In the long run I would like to have an animation like that in original RCT which I like a lot. But in the short run I don't like the image as it is, though I can't quite put my finger down what exactly the problem is or how it could be improved. Any feedback and suggestions welcome :)


doc/data_format.rst
```diff
@@ -1672,73 +1672,83 @@ Gui Sprites
...
+ 240        4     10-     mainmenu_new              Main menu screen: New Game button.
+ 244        4     10-     mainmenu_load             Main menu screen: Load Game button.
+ 248        4     10-     mainmenu_settings         Main menu screen: Settings button.
+ 252        4     10-     mainmenu_quit             Main menu screen: Quit button.
+ 256        4     10-     mainmenu_frames           Main menu screen: Number f of frames in the background animation.
+ 260        4     10-     mainmenu_duration         Main menu screen: Duration of each frame in the background animation in milliseconds.
+ 264      4*f     10-     mainmenu_frame_f          The image for each frame of the main menu's background animation.
+264+4*f     4      2-     texts                     Text of the guis (reference to a TEXT block).
+268+4*f                                             Total length.
+=======  ======  =======  ========================  ==========================================================================================
```
Alberth289346: I wonder if it would be useful to split the main menu items from regular gui sprites, and the former their own block.. While both are guis sprites, the regular sprites are all very small and composed to construct the gui elements. The main menu items are simply big colourful sprites you can click on.

What do you think?

Alberth289346:
I was about to comment there is a logo already, but you found and discarded that :)

All I remember was that they likely had a bunch of savegames that were loaded and running, while displaying the main menu on top of it. Don't remember a logo being displayed at all, although likely my memory is just wrong :)

Noordfrees:
The code is not really polished yet, I'll likely restructure it depending on the feedback. 
I only put the images in the GUI sprites block for convenience for this draft. I'd prefer not to put them in an RCD file at all (it seems pointless IMHO) but instead load the PNG/JPG files from a hardcoded disk location at runtime. We don't yet have any code that does this as far as I can tell so I want to ask whether this would be OK.

Original RCT also has a logo, though less prominent:
grafik
But in any case I think every menu in every game looks better if it features a logo :) 

I had not considered the possibility to display a sequence of savegames. Loading times are brief so this would be possible for us, and coding some showy camera transitions should not be too hard. The downside of it would be that we'd need to include a savegame file which will need resaving whenever we break compatibility (every few weeks currently; quite possibly still once a year or so in the long run). This might become an annoyance. But on the whole I like the idea, shall I implement it like this?

Alberth289346:
> instead load the PNG/JPG files from a hardcoded disk location at runtime

The only thing against that is lack of code for painting such an image to the screen. That is, it's likely much simpler if we can paint everything with the same rendering code. Otherwise you end up with multiple internal data formats for images and multiple rendering functions.

If we would have many large images this may make sense, but for 1 image? (At least I don't see many more such images in the near future?)

If you want a separate file and use the existing rendering code, the solution would be to encode the image as done with RCD files, and then dump that into a file. Depending on the size of the resulting data, it's also feasible to convert the data to a blob of bytes like `uint8 logo[];` and link it into the executable.

My first guess is that it's not worth the trouble of writing new loaders and new rendering code, or perhaps not at this point in time?
Are there large advantages to it that I am missing?


> I had not considered the possibility to display a sequence of savegames.

Actually, one savegame is sufficient even, just jump the camera to different positions every X seconds. Of course you can do both :p (several save games with several camera positions).

> savegame file which will need resaving whenever we break compatibility

Indeed this is a disadvantage. To combat this, the version numbering of the blocks exists. At least in theory, you can update savegame data to a next block version until you are at the current version, and then start the game. In the past I tried to work out a system to implement this, but it didn't get very far due to lack of concrete examples.

Noordfrees:
There we go :) 
I've an idea how to handle the savegame compatibility without too much effort, which I'll implement along with the next feature that changes the saveloading yet again.

Issue #23, opened 2015-03-16
Error message when attempting to build on unowned land needed


GoogleCodeExporter:
```
As indicated in #22, there is currently no indication that you are attempting 
to build on land that you do not own.

Obviously, at some point in the future there will be fences marking this 
boundary, but there should also be some sort of error message.
```

Original issue reported on code.google.com by `CharlesP...@googlemail.com` on 11 Aug 2013 at 11:51

PR #186, opened 2021-03-30
Add error message when attempting to build in unsuited places

Noordfrees:
Re #23 
Currently implemented only for building fixed rides and scenery items.


src/error_gui.cpp
```diff
@@ -17,58 +17,84 @@ enum ErrorMessageWidgets {
...
 }
 
+void ErrorMessageWindow::TimeoutCallback()
+{
+	if (this->timeout_timer_running) {
+		delete this;
+	} else {
+		GuiWindow::TimeoutCallback();
+		if (this->timeout_duration > 0) {
+			this->timeout = this->timeout_duration;
```
sgtbigman, line 78: ~~Is `this->timeout` valid?~~ Disregard :)

src/gamecontrol.cpp
```diff
@@ -259,3 +259,91 @@ void GameModeManager::SetGameMode(GameMode new_mode)
...
+	return true;
+}
+
+/**
+ * If several error messages are applicable, decides which
+ * one is the more important reason to display to the user.
+ * @param older [inout] A reason that existed previously (will be changed if appropriate).
+ * @param other Another reason that is also applicable.
+ */
+void CheckIsMoreImportantReason(StringID *older, const StringID other)
```
sgtbigman: This must have been fun to write :P

sgtbigman (approved):
Overall, looks good to me.

sgtbigman:
Not specific to this PR: @Noordfrees, it's awesome to see someone adding so much to the project. Thank you!

src/gamecontrol.cpp
Alberth289346: You're not checking, you're selecting the most appropriate message.

src/gamecontrol.cpp
```diff
@@ -259,3 +259,91 @@ void GameModeManager::SetGameMode(GameMode new_mode)
...
+	}
+
+	if (*older == GUI_ERROR_MESSAGE_SLOPE || *older == GUI_ERROR_MESSAGE_UNDERGROUND) {
+		if (other != GUI_ERROR_MESSAGE_UNOWNED_LAND) *older = other;
+		return;
+	}
+	if (other == GUI_ERROR_MESSAGE_SLOPE || other == GUI_ERROR_MESSAGE_UNDERGROUND) return;
+
+	*older = other;
+}
```
Alberth289346, line 343: This is going to explode in size, as you basically write down each comparison explicitly.

A simpler approach here is to attach a priority to each message and then pick the message with the highest priority.

src/scenery_gui.cpp
```diff
@@ -43,6 +43,7 @@ class SceneryGui : public GuiWindow {
 	const SceneryType *selected_type;           ///< Currently selected item type.
 	uint8 orientation;                          ///< Current orientation.
 	std::unique_ptr<SceneryInstance> instance;  ///< Instance being placed.
+	StringID build_forbidden_reason;            ///< Reason why we may not place the instance at the given location, if any.
```
Alberth289346: I am guessing we're going to have more of these "compute max priority message"?

If you wrap a class around it, you can move some of the details into that class, eg
```
class BestReason {
public:
   ...
   StringId reason;
   void Reset() { reason = STR_NULL; }
   void UpdateMessage(StringID new_message) { CheckIfMoreImportant(&this->reason, new_message); }
```

graphics/rcd/lang/en_GB.txt
```diff
@@ -70,8 +70,18 @@ strings {
 		QUIT_YES:     "Yes";
 
 		// Error message strings.
-		ERROR_MESSAGE_CAPTION: "Message";
-		ERROR_MESSAGE_SPRITE:  "Failed to draw sprite";
+		ERROR_MESSAGE_CAPTION:        "Message";
+		ERROR_MESSAGE_SPRITE:         "Failed to draw sprite";
```
Alberth289346, line 74: This looks like a weird message. It sounds like a technical thing rather than a user problem.
How is a chief executive of an amusement park going to solve this problem?

Arguably, we should have a different way to deal with these things?

PR #188, opened 2021-04-02
Refactor savegame compatibility handling

Noordfrees:
The current block handling has a few drawbacks if we want to ensure longer-term savegame compatibility. Many `Save()`/`Load()` functions are unaware of the block version. Even if the version is passed down, some are used by *multiple* blocks (e.g. `VoxelObject::Save()` by RIDS and GSTS). And if we increase the overall block version for every small change we'll quickly end up with versions in the hundreds and thousands, and as many different loading paths.
This is changed now by giving every Save/Load function a version number of its own, which is the first number written/read in every such function. This version number is always defined in the same file and increased for every change to the corresponding Save function, so that the multiple Save/Load code paths are self-sufficient; any change that affects saveloading needs to provide compatibility code only in the same place it actually touches.

Further, currently handling invalid data is done by setting the loader's fail message – and then continuing to load even though we know the data is invalid. Depending on the sort of invalid data this may work fine, it may end up in an endless loop, or it may segfault. Now we instead throw an exception on the first sign of invalid data, to prevent loading beyond the problem and to ensure that everything is default-initialized smoothly and safely.

Finally, every savegame is resaved directly after loading to ensure one doesn't need to bother resaving manually. In particular this ensures that the main menu savegame always uses the most recent versions.

---

Loading a master savegame with this branch fails safely. A savegame created with b36f7df can be loaded with 037a27f (which adds a few saveloaded attributes to the `Guest` class) without problem, and the savegame will be updated automatically while doing so.


src/ride_type.cpp
```diff
@@ -775,8 +780,10 @@ void RidesManager::Save(Saver &svr)
 	}
 	svr.PutWord(count);
 	for (uint16 i = 0; i < count; i++) {
-		if (this->instances[i] == nullptr || this->instances[i]->state == RIS_ALLOCATED) continue;
-		this->instances[allocated_ride_indexes[i]]->Save(svr);
+		RideInstance *r = this->instances[allocated_ride_indexes[i]];
+		svr.PutByte(static_cast<uint8>(r->GetKind()));
```
sgtbigman, line 787: Any reason that these two lines were moved outside of `RideInstance::Save`?

sgtbigman (approved):

src/ride_type.cpp
Noordfrees: These two pieces of information are needed by the `RideManager` on loading before it can decide which subclass of `RideInstance` it needs to instantiate. The derived classes save their own version number before calling the base class's Save function though. 
If we kept the old order, the data which is required to instantiate the correct class would be hidden behind several subclass version numbers, which can only be peeled away after the class has been instantiated. So we need to move this data where it can be loaded prior to figuring out which type of ride to allocate.
I hope my explanation is clear… :)

sgtbigman: Makes sense. Thank you :)

src/scenery.cpp
```diff
@@ -257,8 +257,13 @@ bool SceneryInstance::NeedsWatering() const
 	return this->type->watering_interval > 0 && this->last_watered > this->type->watering_interval;
 }
 
+static const uint32 CURRENT_VERSION = 1;   ///< Currently supported version of %SceneryInstance.
```
Alberth289346: Just browsing your changes now, this variable needs a suffix it seems.
Will have a closer look either this evening or tomorrow

As you are changing the save format, I expected some data format document change, but that seems lacking.

sgtbigman:
I'm investigating a segfault that seems to have cropped up in 3b9b6323ed9b491e6b10fbba84c7860bc39f95e7. Are you running into anything locally, @Noordfrees?

Noordfrees:
Nope… what does the segfault's backtrace say? Or how can it be reproduced?

sgtbigman:
Seems to be related to initial load of the mainmenu savegame file. Since it's working for you, it could just be something with my local repo/copy. I'll clean it up and check if that does the trick.

Thanks.

src/scenery.cpp
Noordfrees: Whoops, that doc is seriously outdated by now. Will give it a full overhaul…

src/palette.cpp
```diff
@@ -166,12 +166,17 @@ void Recolouring::AssignRandomColours()
 	}
 }
 
+static const uint32 CURRENT_VERSION = 1;   ///< Currently supported version of %Recolouring.
```
Alberth289346: This one also misses a suffix.

src/shop_type.cpp
```diff
@@ -209,14 +209,19 @@ XYZPoint32 ShopInstance::GetExit(int guest, TileEdge entry_edge)
 
 }
 
+static const uint32 CURRENT_VERSION = 1;   ///< Currently supported version of %ShopInstance.
```
Alberth289346: and this one

src/track_piece.cpp
```diff
@@ -327,8 +327,13 @@ void PositionedTrackPiece::RemoveFromWorld(const uint16 ride_index) {
 	this->piece->RemoveFromWorld(ride_index, base_voxel);
 }
 
+static const uint32 CURRENT_VERSION = 1;   ///< Currently supported version of %PositionedTrackPiece.
```
Alberth289346: and this one

src/fixed_ride_type.cpp
```diff
@@ -321,8 +321,12 @@ void FixedRideInstance::OnAnimate(const int delay)
 	}
 }
 
+static const uint32 CURRENT_VERSION = 1;   ///< Currently supported version of %FixedRideInstance.
```
Alberth289346: and this one

src/gentle_thrill_ride_type.cpp
```diff
@@ -387,8 +387,12 @@ void GentleThrillRideInstance::RecalculateRatings()
 	this->excitement_rating += scenery * t->excitement_increase_scenery;
 }
 
+static const uint32 CURRENT_VERSION = 1;   ///< Currently supported version of %GentleThrillRideInstance.
```
Alberth289346: and another one

I know they are local currently, but the same name for different things means you cannot move them together in one file if you wanted, also `CURRENT_VERSION` doesn't say what thing it gives a version number. They are all 1 now, but in time different `CURRENT_VERSION` will have different values, and then you have no simple way of checking if a `CURRENT_VERSION` variable is actually the correct one,.

Alberth289346:
> The current block handling has a few drawbacks if we want to ensure longer-term savegame compatibility. Many Save()/Load() functions are unaware of the block version. Even if the version is passed down, some are used by multiple blocks (e.g. VoxelObject::Save() by RIDS and GSTS).

I have looked at the current state, and what you're proposing, and I think you're somewhat right and somewhat wrong, so we need to unify our views.

You are correct in not everything having a block number. I wondered why, and came to the conclusion that giving it a separate block number likely didn't make sense. Let's use the `VoxelObject` thing as an example, it's basically a list of voxels. Now imagine we improve it in some way with a new feature. So it's not a voxel object that we have now, it's a new voxel object. Since the voxel object is just storage of information (as far as I can see), that change is only useful if things that use voxel objects change as well. In other words, it seems like block numbers were omitted for data that has a low probability to change without changes in its surrounding management/construction code as well.
With a quick glance this looks like a reasonable reason why there is no version number everywhere currently.

Some of the elements have grown since, eg all `RideInstance` objects are now much more different for different types of rides with your additions. So I agree these could do with a number of their own.
For others I am much less sure. Eg the voxel objects above. Suppose you give them a number. On improvement, they go from number 1 to number 2 (for example), but since other blocks (say block X) use or manage these voxel objects, that code changes too, so those numbers increase too[1]. In addition you get the problem that you have to cross-check version of voxel objects with version of block X since the code of X assumes improved voxel objects.
As such I don't think version increases are limited to a single block here. During loading that probably works, but afterwards you also need the program not to crash on data that is invalid due to having a previous version.

[1]: Obviously, you only need to change the block of X if the stored data actually changes. That however makes things somewhat more complicated, because you then get a single version of block X supporting several different VoxelObject versions, depending on the point in time that you copied the X code. That is, the X version number doesn't say if it can use voxelobjects version 1 or 2 any more.

I agree we should have more clarity in how different kinds of data are mixed in a single block. Not sure how to do that currently.


> This is changed now by giving every Save/Load function a version number of its own

Giving each ride instance the same version number is somewhat blunt, as it's always the same for every ride of the same type. More importantly, with this patch we now have 2 version number systems, one with block names and sizes and stuff, and one 'arbitrarily' in the middle of the data stream. I'd prefer that we stick with one system.
You could for example make ride instance blocks (either generic or separate blocks for different ride instance types), and write all rides of a single type into a single block.

For ride instances, I'd say this is a fair evolutionary step, they really have grown. for other smaller data, I am much less sure there is any benefit due to the coupling of version numbers of different blocks in the code that manipulates the data. However, if you see benefits, I can live with them, providing you don't invent new version number systems, one system is complicated enough I would say. [Also, as these new version numbers are inside other blocks, the documentation requires you to still update each and every use in the existing blocks.]

> Now we instead throw an exception on the first sign of invalid data, to prevent loading beyond the problem and to ensure that everything is default-initialized smoothly and safely.

Note sure we need an exception for this, but abort on bad data is fine to me.

> Finally, every savegame is resaved directly after loading to ensure one doesn't need to bother resaving manually. In particular this ensures that the main menu savegame always uses the most recent versions.

Why? If you can load the old version once, you can do it again. No need to update the file at all. In fact, your solution may even destroy save games.
Suppose you make a new update with a new version. You run the program, load a save game, and then find out you missed a case in the load upgrade process. Now you're stuck with a file in the new format that is broken. No way to undo the changes you made. Are you going to invent ad-hoc recognition code for fixing such broken savegames???
On the other hand, if you loaded the file, and didn't write a new version, you can fix the load procedure, the game can be loaded and converted again, and you can continue playing that game.
If the only reason is a handful of files that only exist for the front-page of the game, make an "--autosave-upgraded-version" option or whatever (and even that isn't needed if you can load old files).
Don't destroy data without the user telling you he/she really wants to do that. Computers are fast but not smart, despite advertising that says otherwise.

src/gentle_thrill_ride_type.cpp
Alberth289346: I stopped searching at this point, please do a global search to find all  of them.

Noordfrees:
> Some of the elements have grown since, eg all RideInstance objects are now much more different for different types of rides with your additions. So I agree these could do with a number of their own.
> For others I am much less sure. Eg the voxel objects above. Suppose you give them a number. On improvement, they go from number 1 to number 2 (for example), but since other blocks (say block X) use or manage these voxel objects, that code changes too, so those numbers increase too[1]. In addition you get the problem that you have to cross-check version of voxel objects with version of block X since the code of X assumes improved voxel objects.

`VoxelObject` is the base class of `Person` and `CoasterCar` (and possibly more things in the future, who can tell). If I make a change that should affect e.g. Persons but not Cars, then this change should not go into the VoxelObject class but into Person code, and saveloaded by Person rather than VoxelObject. If I make a change to VoxelObject, then because I *want* this change to affect *all* subclasses of VoxelObject. So this is exactly the behaviour I consider desired.

> As such I don't think version increases are limited to a single block here. During loading that probably works, but afterwards you also need the program not to crash on data that is invalid due to having a previous version.

That problem can be solved by simply assigning reasonable default values during loading. This is applicable in nearly all cases; and in those few where it's not, I doubt that any system – neither the old one nor the new one – would allow us to keep supporting the older savegame versions.

> [1]: Obviously, you only need to change the block of X if the stored data actually changes. That however makes things somewhat more complicated, because you then get a single version of block X supporting several different VoxelObject versions, depending on the point in time that you copied the X code. That is, the X version number doesn't say if it can use voxelobjects version 1 or 2 any more.

That's… actually the core point of this change. To decouple the several layers of Save functions within each block. You could think of the various saveload codes as nested sub-blocks. A block SCNY at version 1 contains certain SCNY-specific data, and then N times the data layout of a SceneryInstance. Each of these in turn contains first a SceneryInstance version number and then the SceneryInstance data layout specific to that version.

> Giving each ride instance the same version number is somewhat blunt, as it's always the same for every ride of the same type. More importantly, with this patch we now have 2 version number systems, one with block names and sizes and stuff, and one 'arbitrarily' in the middle of the data stream. I'd prefer that we stick with one system.
> You could for example make ride instance blocks (either generic or separate blocks for different ride instance types), and write all rides of a single type into a single block.

Previously, all saveloading code is centred around the idea of having multiple blocks with a fixed layout which incompasses a lot of data. Let's call this a **block** layout. This branch uses an **encapsulated** layout instead. We have many **patterns** of data, each of which is self-sufficient with its own version number. Every pattern can in turn contain more patterns. The layout of these *nested patterns* is hidden from the *parent pattern*; parent patterns can use nested patterns without knowing or caring about the implementation details (i.e., the layout) of their nested patterns.
Code-wise, every pattern is represented by exactly one `Save()` and one `Load()` function, and each such function represents exactly one pattern. This allows the saveloading functions to call other saveloading functions (which represent the nested patterns) without knowledge about the nested functions' version numbers and data layout.
The blocks are, in this terminology, ordinary patterns, with the exceptions that these are not nested in any other pattern and that every non-block pattern in nested (directly or indirectly) in a block pattern.

> For ride instances, I'd say this is a fair evolutionary step, they really have grown. for other smaller data, I am much less sure there is any benefit due to the coupling of version numbers of different blocks in the code that manipulates the data. However, if you see benefits, I can live with them, providing you don't invent new version number systems, one system is complicated enough I would say. [Also, as these new version numbers are inside other blocks, the documentation requires you to still update each and every use in the existing blocks.]

We might think now that there are data structures so trivial that we will likely never, ever change their layout (or that, if we do, it will likely be accompanied by big changes in the surrounding code). But if we are so certain of it now and in a few years need to make such a change after all, this "optimization" will backlash badly. I've seen that before, and it can turn what should be a ten-lines-change into the full rewrite of two files. 
It is safest to give every pattern a version number. Not to mention that this is required for consistency with the system as described above; otherwise there would be patterns without a version number of their own, which are therefore not truly encapsulated from the parent pattern.

> > Now we instead throw an exception on the first sign of invalid data, to prevent loading beyond the problem and to ensure that everything is default-initialized smoothly and safely.

> Note sure we need an exception for this, but abort on bad data is fine to me.

Throwing and catching the exception allows us to recover and create a default-initialized game world instead of loading. In the future we can also show an error message "Your savegame is damaged because of %s, it is not compatible with this version of FreeRCT". Allowing a program to crash, in my opinion, should be allowed only if it is completely impossible to avoid it.

> Why? If you can load the old version once, you can do it again. No need to update the file at all. In fact, your solution may even destroy save games.
Suppose you make a new update with a new version. You run the program, load a save game, and then find out you missed a case in the load upgrade process. Now you're stuck with a file in the new format that is broken. No way to undo the changes you made. Are you going to invent ad-hoc recognition code for fixing such broken savegames???
> On the other hand, if you loaded the file, and didn't write a new version, you can fix the load procedure, the game can be loaded and converted again, and you can continue playing that game.
> If the only reason is a handful of files that only exist for the front-page of the game, make an "--autosave-upgraded-version" option or whatever (and even that isn't needed if you can load old files).
> Don't destroy data without the user telling you he/she really wants to do that. Computers are fast but not smart, despite advertising that says otherwise.

I'll implement such an option, but I don't quite agree with the first argument.
This is true now, and probably for some time more, but certainly not over a period of five years from now. If we keep supporting loading older versions for ages, the size of the loading code will quickly multiply. We will want to drop support for ancient versions occasionally, e.g. drop support for version X if version X+1 has been introduced more than one year or half a year ago. Users can be expected to now and then resave old games they want to keep. Automatically resaving on load would be very convenient here (and for the main menu savegame as well btw). 
And the argument that this might destroy the file if there's a bug in the loading code applies only to developers who actually modify the saveloading code (and those users brave enough to test development branches), but not when playing master. We may hope that in the long run there will be far more players than developers, so it might make sense to enable overwriting by default and add the command-line switch (or config file entry) to disable it.

Alberth289346:
> This branch uses an encapsulated layout instead. We have many patterns of data, each of which is self-sufficient with its own version number. Every pattern can in turn contain more patterns. The layout of these nested patterns is hidden from the parent pattern; parent patterns can use nested patterns without knowing or caring about the implementation details (i.e., the layout) of their nested patterns.

You are very confident that you never mess up numbers, sizes, or offsets. Without synchronization points with eg block names and sizes, you never know if a value is a true version, or just random data from some other block, It may cause failure further down the file, but that may also not happen. [Block names and sizes are also not 100% air tight, but iirc, you cannot read beyond a block, and knowing myself you can also not skip reading the last part of the block, so if an error happens here, you should get pretty good pointers where the problem is.

"not caring" may be true for sequential loading, but with just version of an inner pattern, it's not possible to extract the data of the outer pattern only, you need at least the size of the inner patterns to do that. You cannot find all instances of the internal patterns either, for the same reason.

Blocks are designed so you can jump from block to block quickly, checking eg the blocks used, and their version without ever knowing what they store. You know where it starts, you have a name and a size, so you can find the next block without needing details of block internals. Version number isn't even needed to do that (although it's nice to have there, and useful if you do need to read the internals).

The second point is that in the current block system, the version number of the block completely defines the block structure. Thus the version number is enough to decide whether you can load the block without even looking inside it. In the pattern idea, this no longer holds. Inner patterns may change without being visible in the outer block version. Worse, they may change size and position if you nest patterns. As a result, to decide whether you can load a block, you must recursively parse all patterns inside it. That is, check all of them. This in turn requires that this tester-program knows the complete layout of all patterns that exist (since layout of a pattern is the only way to find the next (nested) pattern), and every version of every pattern can have different size or position of its inner patterns.

The nested pattern idea thus kills the quick and simple way to jump over the blocks and do a quick scan, find a specific block, etc etc, basically custom querying of data from a file without having to know the size and offsets of all data of all versions of all patterns. I think that is a pity, and I am far from sure we want to discard that option.

> We might think now that there are data structures so trivial that we will likely never, ever change their layout (or that, if we do, it will likely be accompanied by big changes in the surrounding code). But if we are so certain of it now and in a few years need to make such a change after all, this "optimization" will backlash badly. I've seen that before, and it can turn what should be a ten-lines-change into the full rewrite of two files.

If a block changes, you change the version number, period. In other words, you should never change voxel object storage without giving each block that contains it a new version. If you fail to do that, it's broken. As said, I agree that this dependency is not entirely clear currently and should be made more explicit.
A simple way to do this would be to give all inner patterns a version number in the code, and assert this number matches with the version of the surrounding pattern. If you statically assert that, you cannot change the version of the inner pattern without also fixing the surrounding pattern, it simply won't compile if you forget it. That would at least force consistency in version changes, I think.

---
I can see the benefits of inner and outer patterns and nesting them. It's a nice idea. For voxel objects your idea might be somewhat tricky, as conceptually with voxel objects being a base class, voxel objects don't quite exist on their own. (It might be a good idea applied to the wrong case here.)

What worries me though is that you put inner patterns inside an outer pattern without being able to find the inner patterns in the sea of bytes of the outer pattern. You must have full information about the outer pattern to find them, and then get from the outer pattern full information what inner pattern to expect, and from the inner pattern data the version and its full information to decide about nested inner patterns and size.
That's a whole lot of jumps and detailed information retrieval with very tight dependencies. Everything must be right, or you're lost in the byte sea, no way to distinguish inner and outer pattern data from each other. So if we ever make a mistake somewhere there, it's going to be a long search for the culprit.

To avoid such problems, we either don't nest data like that (current solution, probably extended with the version asserts above), or we find a way to have nested blocks/patterns, while saving enough meta-data so you can still write a simple program that can jump from block to block (pattern to pattern) without needing full knowledge of patterns internals.

For the latter, it would be enough if the direct inner patterns are at a fixed position inside the outer pattern, with the inner patterns also storing their name and size (and version or you still can't dig inside the inner pattern if you need to). The outer pattern must store where the inner patterns can be found, so you can distinguish between inner patterns and data of the surrounding pattern.
One solution is to have a block like we have now, with the following changes:
- All patterns are stored like blocks, with a name, a version, and a size. In addition they also have an offset field.
- All inner patterns are stored sequentially from the start of the outer block.
- Data of the outer block follows after the last inner pattern.
- The offset field of the outer block identifies the total size of all contained inner patterns (and thus also makes it possible to know where the inner patterns end, and where the outer block data starts).

Of course you can also do it the other way around, first have outer block data and then the inner pattern blocks, not sure what is most convenient for loading (save happens only once, so seems less of a concern here, unless it becomes very complicated to save a file.)

How does forced version management, or a nested pattern thing like above look to you?

---
> A block SCNY at version 1 contains certain SCNY-specific data, and then N times the data layout of a SceneryInstance. Each of these in turn contains first a SceneryInstance version number and then the SceneryInstance data layout specific to that version.

As we never write different versions for the same type of patterns in one save (I see no valid use case for that), you're still writing lots of duplicate data. Arguably, this isn't bad, as disk space is cheap.

> Throwing and catching the exception allows us to recover and create a default-initialized game world instead of loading.

You don't necessarily need an exception for such behavior, you can simply return a boolean "failed", or in case of the sprite loader, a non-null string describing the problem. I only wanted to say, throwing an exception is not the one and only way to achieve this.


> This is true now, and probably for some time more, but certainly not over a period of five years from now. If we keep supporting loading older versions for ages, the size of the loading code will quickly multiply. We will want to drop support for ancient versions occasionally, e.g. drop support for version X if version X+1 has been introduced more than one year or half a year ago. Users can be expected to now and then resave old games they want to keep.

Expiration of old versions is a separate problem imho. In my experience, code doesn't multiply, it grows linearly with each new version (you get on average around Y lines additional code for each new version, a new version doesn't take more lines because you have lots of previous versions, since version upgrades can be applied sequentially).
I agree we will want to drop old versions at some point. Not sure how to do that though. Simplest is perhaps by major version. If we have major version N, then we support loading saves of major version N-1, but not N-2 or older. (Obviously, we can support more than one previous major version if we like, that's a policy decision.)

Your autosave feature reduces the problem, but doesn't solve it. I may have a set of very important save games stashed on an USB stick lying on my desk as backup and then the disk crashes after several years (and I didn't make a new backup). Alternatively, I may have gotten bored of playing freerct,  played a different game for a few years, and then return to this game. Or I may have downloaded some old saves from the Internet.

In other words, you won't convert every save game that exists soon enough. There will always be cases where conversion isn't possible anymore, as soon as you start throwing old conversion code away.


> Automatically resaving on load would be very convenient here

You are skipping a few steps here. I said don't do that without user permission. That permission can take many forms. It includes having to press a button, but other forms are asking the user with a popup "I converted the file, do you want to save it now?", and many other forms, upto an option in a configuration somewhere "automatically save converted save files (yes/no)".
The user has to say it's allowed somewhere, that's alll I ask. Make it possible to opt-in rather than decide things for the user without explicit permission in some way.

> And the argument that this might destroy the file if there's a bug in the loading code applies only to developers who actually modify the saveloading code (and those users brave enough to test development branches), but not when playing master.

You're over-confident here. I developed in OpenTTD for some years, and I found bugs in code that existed years in master already (named trunk at that time) before someone stepped on it. I also found bugs that got introduced in the very first commit in the repository, so that bug was around 20 years old. Sure developers are the first and the most likely to notice problems here, but saying you can test enough that you won't ever screw up in released code is a claim you cannot make true unless you never release code.

Noordfrees:
> You are very confident that you never mess up numbers, sizes, or offsets. Without synchronization points with eg block names and sizes, you never know if a value is a true version, or just random data from some other block, It may cause failure further down the file, but that may also not happen. [Block names and sizes are also not 100% air tight, but iirc, you cannot read beyond a block, and knowing myself you can also not skip reading the last part of the block, so if an error happens here, you should get pretty good pointers where the problem is.
> 
> "not caring" may be true for sequential loading, but with just version of an inner pattern, it's not possible to extract the data of the outer pattern only, you need at least the size of the inner patterns to do that. You cannot find all instances of the internal patterns either, for the same reason.
> 
> Blocks are designed so you can jump from block to block quickly, checking eg the blocks used, and their version without ever knowing what they store. You know where it starts, you have a name and a size, so you can find the next block without needing details of block internals. Version number isn't even needed to do that (although it's nice to have there, and useful if you do need to read the internals).
>
> The second point is that in the current block system, the version number of the block completely defines the block structure. Thus the version number is enough to decide whether you can load the block without even looking inside it. In the pattern idea, this no longer holds. Inner patterns may change without being visible in the outer block version. Worse, they may change size and position if you nest patterns. As a result, to decide whether you can load a block, you must recursively parse all patterns inside it. That is, check all of them. This in turn requires that this tester-program knows the complete layout of all patterns that exist (since layout of a pattern is the only way to find the next (nested) pattern), and every version of every pattern can have different size or position of its inner patterns.
> 
> The nested pattern idea thus kills the quick and simple way to jump over the blocks and do a quick scan, find a specific block, etc etc, basically custom querying of data from a file without having to know the size and offsets of all data of all versions of all patterns. I think that is a pity, and I am far from sure we want to discard that option.
>
> […]

I strongly disagree with that. The current blocks implementation does NOT allow skipping over blocks, *especially* not without regarding the version number.

Consider the Guests block. It contains at some point the number of guests, followed by the data of as many guests as specified by that number. How would you skip this block without reading this number first? And if we might add more data between the block start and this number in a future version, you'll need to distinguish between the various versions to determine the offset of the numbers that determine the length of the block, and therefore also the starting position of the next block! 
There are plenty more examples, the Rides and Voxelstack blocks for instance. 

If block-scanning would work now, this would be a drawback; but as it is, we are *not* losing any previously existent functionality with the pattern approach.

I can certainly add some magic strings at the begin and end of every pattern though if you like.

> You don't necessarily need an exception for such behavior, you can simply return a boolean "failed", or in case of the sprite loader, a non-null string describing the problem. I only wanted to say, throwing an exception is not the one and only way to achieve this.

Would also work, but then we'd need to pass the `false` return value up along all functions to avoid making any load-time decisions based on invalid data. We'd have to change every single load call of a sub-pattern like
``` diff
- x->Load(ldr);
+ if (!x->Load(ldr)) return false;
```
An exception effectively has the same result with far fewer codechanges – and without the risk that the check might be forgotten in some places.

Alberth289346:
> Consider the Guests block. It contains at some point the number of guests, followed by the data of as many guests as specified by that number. How would you skip this block without reading this number first?

That's wrong then. The definition says the header has a size field equal to the size of the block.

Noordfrees:
> > Consider the Guests block. It contains at some point the number of guests, followed by the data of as many guests as specified by that number. How would you skip this block without reading this number first?
> 
> That's wrong then. The definition says the header has a size field equal to the size of the block.

Where does the documentation say that? 
And when we say "documentation", we're both talking about https://github.com/FreeRCT/FreeRCT/blob/master/doc/savegame.rst, right?

Alberth289346:
AARGHH!

Spend a whole day discussing with the wrong format in my head :(

Alberth289346:
Having fixed patterns like the existing blocks would be good, so you don't introduce new block separation conventions, imho

Noordfrees:
I renamed the `savegame.rst` to `savegame_history.rst` with only small changes, and wrote documentation for the pattern approach as it is implemented now in `savegame.rst`. I hope it is clear how I defined the patterns and their nesting relations there.

doc/savegame.rst
```diff
@@ -27,25 +27,42 @@ Introduction
...
+          Documentation for older format versions can be found in the file "savegame_history.rst".
+
 The file starts with a file header to identify the file as being a save game.
-After the file header come data blocks of game elements that are stored.
+After the file header come data patterns of game elements that are stored.
+
+All data in the file is organized in **patterns**. Each pattern starts by
+specifying a version number followed by the pattern's data layout as
+detailed below. A pattern can contain other nested patterns. A pattern
+which is not nested in any other pattern is called a **block**.
```
Alberth289346: No header and footer data like you proposed yesterday? What happened to it?

doc/savegame.rst
```diff
@@ -27,25 +27,42 @@ Introduction
...
+   ?       ?     10-     Current basic world block.
+   ?       ?     10-     Current financial data.
+   ?       ?     10-     Current weather block.
+   ?       ?     10-     Rides block.
+   ?       ?     10-     Scenery block.
+   ?       ?     10-     Guests block.
+   ?       ?     10-     Staff block.
+   ?       ?     10-     Inbox block.
+   ?       ?     10-     Current random number block.
    ?                     Total length of the save file.
```
Alberth289346: Almost everywhere, the total size isn't fixed, so maybe just drop the "total size" line at all?

Arguably the offset column could be deleted too, but that may be useful enough to keep.

Would it be useful to distinguish fields from patterns? Eg make the patterns bold, so a reader knows it's defined else where. Ideally it would be a link, but that may be too much work in REST format.

PR #189, opened 2021-04-02
Switch person graphics

### LINKED ISSUE #72: Switch person graphics

Noordfrees:
Closes #72
… with two small drawbacks. The 64-px sprites are so tiny I can barely see them, so I'm using the 128-px sprites, but they're pretty large. And the animation has a varying number of frames for the four orientations (1×23, 1×25, 2×24) so I had to clip all of them to only 23 frames each. Do we want this anyway?


Alberth289346:
I am fine with not having them.
Keeping the animation support is likely useful though. Don't we have animations already elsewhere yet?

I vaguely remember such a thing, but I'd have to look it up.

Noordfrees:
We have two kinds of animations: Timed animations (TIMA block in the RCD files), which are used by fixed rides and scenery; and ANIM/ANSP blocks, which are used by Persons (currently only the single-frame pillar planet guests). Both types contain various additional information that is needed by the different usage situations, so merging them into one format would not be useful.

I've reverted the change to the RCD file to keep using the pillars, but the codechange will be useful for when we have other animated guest graphics.

---

Below are two kinds of human graphics I made for another game a while ago. If you happen to like any of these (regarding the style as well as stylish consistency) I could adapt them for FreeRCT.
grafik
grafik

Alberth289346:
Persons look too detailed to me, pillar people are just a handful of pixels (less than 10 high iirc, but it was quite a few years ago).

I think the most important is that they are colourful, so you can clearly see them on rides if we get the graphics for it.

Noordfrees:
I have tried out some adjustments to the colours and then pasted my models into the coaster car:
grafik grafik
(the whitish spots on the trains are due to me messing up the 8bpp palette)

IMHO they are clearly visible like this (and the clothes will also make good recolouring targets), what do you think?

Alberth289346:
Looks great to me, more than enough to develop the code for guests in the ride.

I guess you need to draw 2 additional sprites just after painting the car (1 sprite for each bench) to add the pixels of the guests, and indeed some recolouring of each guest. That would give the same limitations (probably) of the original (each bench was either full or empty).
Hmm, if you could recolour to full transparency of a guest, you'd get the option of having one guest at a bench. EDIT: Arguably, it may be simpler to have a sprite with 1 guest :)

It does need some trickery in the order, you want to add guests from back to front in depth, so sometimes you first have to add the front row and at other orientations you first have to add the back row.
Ideally, the graphics should store all this information I guess.

Noordfrees:
Here are the new guests. In this PR I'll add only the walk animations, and the onride coding and graphics in a follow-up. I wonder whether these graphics are a bit too tall, should I make them smaller?

For the on-ride guest images I already have some magic in mind that involves adding a lot of new information to the recolour masks (and/or possibly a second mask in case I run out of indices).

Alberth289346:
If you build a road over another road, persons should be able to walk under the bridge. Also, and perhaps more limiting, they should be able to use the ride entrances and exits.

For reference a few enlarged image details from the original. People seem around 15-16 pixels high
Screenshot from 2021-04-07 21-26-38
Screenshot from 2021-04-07 21-28-02
Screenshot from 2021-04-07 21-29-21
Screenshot from 2021-04-07 21-30-54
Screenshot from 2021-04-07 21-31-23

Alberth289346:
I think the important part now is getting the functionality, we have loads of graphics to add anyway to get a full game.

Let me know when you're happy with the result.

Noordfrees:
Comparing the proportions to original RCT is of course a good idea… I had not had in mind how *tiny* their guests are in relation to the entrance buildings. I think I found a good size now, so from my side this is ready for merging :)

PR #191, opened 2021-04-23
Draw guests on coasters

Noordfrees:
grafik


src/coaster.cpp
```diff
@@ -374,8 +403,10 @@ void CoasterTrain::SetLength(const int length)
 	this->cars.clear();
 	this->cars.resize(length);
 	for (CoasterCar &car : this->cars) {
+		car.owning_train = this;
 		car.front.car_type = this->coaster->car_type;
 		car.back.car_type = this->coaster->car_type;
+		car.front.owning_car = car.back.owning_car = &car;
```
Alberth289346: Code usually doesn't daisy-chain assignments for clarity reasons.

src/coaster.h
```diff
@@ -53,13 +53,28 @@ class CarType {
...
+	 * Retrieve a guest overlay sprite at a given \a pitch, \a roll, and \a yaw orientation.
+	 * @param pitch Required pitch of the car.
+	 * @param roll Required roll of the car.
+	 * @param yaw Required yaw of the car.
+	 * @param slot Seat number of the guest to overlay.
+	 * @return The overlay in the requested orientation, if available.
+	 */
+	const ImageData *GetGuestOverlay(uint pitch, uint roll, uint yaw, uint slot) const
+	{
+		return this->guest_overlays[slot * 16u*16u*16u + ((uint)(pitch & 0xf) | ((uint)(roll & 0xf) << 4) | ((uint)(yaw & 0xf) << 8))];
```
Alberth289346: The more modern style of these casts is to use `std::static_cast<T>(value)`, although I would guess there are loads of old-style casts in the code too.

src/map.cpp
```diff
@@ -155,6 +155,17 @@ VoxelObject::~VoxelObject()
  * @return Sprite to display for the voxel object.
  */
 
+/**
+ * Get the overlay sprite(s) to draw for the voxel object.
+ * @param sprites Sprites at the right size for drawing.
+ * @param orient Direction of view.
+ * @return Sprites to overlay for the voxel object, with their respective recolourings.
+ */
+std::vector<std::pair<const ImageData*, const Recolouring*>> VoxelObject::GetOverlays(const SpriteStorage *sprites, ViewOrientation orient) const
```
Alberth289346: It's always tempting to stick an available solution like `pair` into there when you are just outside standard types. The cost of that is a long type (which is mostly fixable with a `typedef`, so not much of a problem), and at every spot where you need to deal with this data, non-intuitive `.first` and `.second` field names, and no way to attach some intelligence into the `pair` by adding a method to it, leading to shorter and more readable code.

I am not sure if using a self written class here instead of `pair` improves things, but in my experience the threshold for that is pretty low in most cases. Please consider.

utils/GenerateSpritesheet.java
```diff
@@ -5,40 +5,43 @@
...
  * Let's assume you've made an animation with `f` frames for a ride with a size of `x` × `y` tiles.
  * The base images, all of which are of size `w` × `h`, are located in `src_dir`.
  * The images are called "PREFIX0000.png" through "PREFIX9999.png", where `PREFIX` may be different for each of the four rotations
- * (let's call the prefixes `se`,`ne`,`nw`,`sw` in this order). The number of the animation's starting frame is `idOff`.
+ * (let's call the prefixes `se`,`ne`,`nw`,`sw` in this order). The number of the
+ * animation's starting frame is `idOff_<p>` for each of the four prefixes <p>.
  * To generate the spritesheet and save it as `output_file`, run
- *    java GenerateSpritesheet src_dir output_file idOff se ne nw sw f x y w h
+ *    java GenerateSpritesheet src_dir output_file idOff_se se idOff_ne ne idOff_nw nw idOff_sw sw f x y w h true
+ * The last, optional parameter can be set to `false` to disable intelligent calculation of frame indices
```
Alberth289346: It looks like you're starting to run into limitations of the command line, doesn't it?

I was always amazed by the large number of different ways to index sprites at sheets :)

utils/GenerateSpritesheet.java
```diff
@@ -5,40 +5,43 @@
...
+ * (useful when (ab)using this tool to generate spritesheets for a different purpose than the one stated above).
```
Alberth289346: You're entering dangerous territory here, in a year or two you won't remember exactly what this "intelligent calculation was, or if you do, you won't remember what the other behavior is.

To be fair, I never documented the sprite indexing in the rcd input files either, so I won't say you have to document it, but please be careful what you don't explain.

utils/GenerateSpritesheet.java
```diff
@@ -16,17 +16,27 @@
...
  * To generate the spritesheet and save it as `output_file`, run
- *    java GenerateSpritesheet src_dir output_file idOff_se se idOff_ne ne idOff_nw nw idOff_sw sw f x y w h true
- * The last, optional parameter can be set to `false` to disable intelligent calculation of frame indices
- * (useful when (ab)using this tool to generate spritesheets for a different purpose than the one stated above).
+ *    java GenerateSpritesheet src_dir output_file idOff_se se idOff_ne ne idOff_nw nw idOff_sw sw f x y w h TIMA
+ *
+ * The last parameter declares which indexing algorithm to use. The indexing algorithm decides at which coordinates
+ * in the resulting spritesheet each input image should be located. Available indexing algorithms are:
+ * - TIMA: Arranges the individual images in the way expected by the TIMA and FSET node generators.
+ * - PLAIN: Just put one sprite after the other, without considering rotations or other fancy processing.
```
Alberth289346: If you stick that in a Java doc with the constants, an IDE will display that whenever you come near the value, giving your efforts much more impact.

src/coaster.cpp
```diff
@@ -250,15 +250,15 @@ const ImageData *DisplayCoasterCar::GetSprite(const SpriteStorage *sprites, View
...
-std::vector<std::pair<const ImageData*, const Recolouring*>> DisplayCoasterCar::GetOverlays(const SpriteStorage *sprites, ViewOrientation orient) const
+VoxelObject::Overlays DisplayCoasterCar::GetOverlays(const SpriteStorage *sprites, ViewOrientation orient) const
 {
-	std::vector<std::pair<const ImageData*, const Recolouring*>> result;
+	Overlays result;
 	if (this->owning_car == nullptr) return result;
 	for (int i = 0; i < this->car_type->num_passengers; i++) {
 		if (this->owning_car->guests[i] != nullptr) {
-			result.push_back(std::make_pair(
+			result.push_back(Overlay {
```
Alberth289346, line 259: In C++11 iirc, there is `emplace_back` which takes the parameters of the constructor of the vector element, and does what you do here (construct an object from the parameters, and add it to the vector), except the new element is not copied.
I don't think we actually use that in the code yet, but it could be worth considering at some point.

Alberth289346:
I think it's fine in the current state, not sure if you want to further improve on my suggestions, or just take them aboard for a next time.
Anyway, I approve this patch, feel free to merge it when finished with it.

And really great work!

4K sprites does feel a bit excessive, but it's the only sane solution :)

Noordfrees:
My compiler does not seem to like `emplace_back` here, perhaps because I'm using an initializer list instead of an explicitly declared constructor, so I left that out for now.

Thanks for the review :) 

And if 4k sprites seem excessive, just wait until I implement this for the gentle/thrill rides as well – the shipswing can hold 4 times as many people as a coaster car :P

PR #194, opened 2021-04-26
Staff Management

### LINKED ISSUE #60: Initial staff (employee) management.
```
(Please attach the patch in unified diff format.)
What are the major things which your patch changes / adds?
Staff management window, employee creation (hire), info window, and deletion 
(sack employee).
Against what version did you make your patch?
Rev 1332

Please provide any additional information below.
Patch is located here: https://gist.github.com/sgtbigman/5b1413cb6ddef3c86b73
Plan to attach patch here after approval.
```

Original issue reported on code.google.com by `SGTBIG...@gmail.com` on 5 Nov 2014 at 7:05

Noordfrees:
This adds a GUI to hire and dismiss staff members. You can use the toolbar button to open the staff management window, and you can click on any staff member to open his info window.
(The toolbar is getting pretty crowded, we'll need to switch to pictorial buttons soon.)

I also added simple classes for the remaining three staff types, but they don't do anything special yet except walk around and demand a salary for it.

I guess this closes #60


src/person.h
```diff
@@ -263,14 +289,100 @@ class Mechanic : public Person {
...
+
+	void ActionAnimationCallback() override
+	{
+		NOT_REACHED();  // Entertainers don't have action animations.
+	}
+
+	const Money &GetSalary() const override
+	{
+		return SALARY;
+	}
```
Alberth289346: I don't remember that the game allowed changing the salary, so it's a fixed number I guess.

An option here is to handle salary in the base class, as it's the same code for all types of staff. Simplest may be to have the salary constant be passed into the base class in its constructor.

src/person.h
```diff
@@ -263,14 +289,100 @@ class Mechanic : public Person {
...
+	/** What a handyman is doing right now. */
+	enum Activity {
+		ACTIVITY_NONE,      ///< Wandering around.
+		ACTIVITY_WATER,     ///< Watering a flowerbed.
+		ACTIVITY_SWEEP,     ///< Sweeping the path.
+		ACTIVITY_EMPTY_NE,  ///< Emptying the bin on the north-east edge of the current tile.
+		ACTIVITY_EMPTY_SE,  ///< Emptying the bin on the south-east edge of the current tile.
+		ACTIVITY_EMPTY_SW,  ///< Emptying the bin on the south-west edge of the current tile.
+		ACTIVITY_EMPTY_NW,  ///< Emptying the bin on the north-west edge of the current tile.
+	};
```
Alberth289346, line 347: Something "Handyman" in these names is likely useful, and/or rename the "none" (perhaps "wander"?) activity, as that is very generic. 
Another way to have a 'handyman' in there is to use an enum class, where you are forced to also list the enum name, eg `Activity::ACTIVITY_NONE`. Then you always have 'activity' twice in the current implementation, so there is room for the staff name, or dropping of one activity.
(I'd aim for something like `HandymanAction::WANDER`, I think.)

[enum classes does make merging all activities into one spot, eg the base class, more difficult due to the increased correct typing that is enforced, so it may not fly here?]

src/person.h
```diff
@@ -263,14 +289,100 @@ class Mechanic : public Person {
...
+	void Save(Saver &svr);
+
+	void ActionAnimationCallback() override;
+
+	const Money &GetSalary() const override
+	{
+		return SALARY;
+	}
+
+	Activity activity;  ///< What we're doing right now.
```
Alberth289346: 'we'?  'the handyman', or 'the staff member' seems more appropriate.

src/person_gui.cpp
```diff
@@ -125,17 +126,111 @@ void GuestInfoWindow::OnChange(ChangeCode code, uint32 parameter)
...
+	if (number == SIW_DISMISS) _staff.Dismiss(this->person);  // This also deletes this window.
+}
+
+void StaffInfoWindow::OnChange(ChangeCode code, uint32 parameter)
+{
+	switch (code) {
+		case CHG_DISPLAY_OLD:
+			this->MarkDirty();
+			break;
+		case CHG_GUEST_COUNT:  // This signifies that our person was deleted. Close the window.
```
Alberth289346: Bit weird constant name?

Alberth289346:
good step

Issue #51, opened 2015-03-16
Ability for guests to walk on some, but not all, pathless lands


GoogleCodeExporter:
```

What are the major things which your patch changes / adds?
If you delete a path below a guest, the guest should continue walking for some 
land slopes.

Against what version did you make your patch?
Revision 1194

Please provide any additional information below.
The following land slopes are not included:

These types are where all corners are raised except for the "not" corner
not north
not south
not east
not west

steep north
steep south
steep east
steep west
```

Original issue reported on code.google.com by `jgi...@my.smccd.edu` on 26 May 2014 at 6:27
- Blocking: #38

Attachments:
- [person.patch](https://storage.googleapis.com/google-code-attachments/freerct/issue-51/comment-0/person.patch)

Issue #38, opened 2015-03-16
Crash when removing a path below a guest


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. build a path upwards
2. let the guest going there
3. remove path under the guest

What is the expected output?
guest fall down to tileground

What do you see instead?
programmcrash

NOT_REACHED triggered at line 167 of 
/home/niels/develop/git/freerct-read-only/src/person.cpp


```

Original issue reported on code.google.com by `vatriani.nn` on 11 Jan 2014 at 10:21
- Blocked on: #51

ghost:
I'm making this comment mostly in order to remind myself to finish fixing this issue, since it shouldn't be that hard to finish. :P

Anyways, my current code, as I have stated before, allows guests to walk on ground tiles. The problem that arises is that 'steep' slopes occupy two voxels, which means that guests can hover above or be partially hidden behind these steep slopes.

I believe the best way to handle this will be to have the guests check which 'corner' they are in (N, E, S, W), based on `pix_pos.x` and `pix_pos.y`, and whether or not the corner is 'raised'. In doing so, we can then place the guest in the correct voxel (if they are not currently in the right one).

I'll continue updating this issue as time permits.

PR #195, opened 2021-04-27
Implement handymen watering the flowerbeds

Noordfrees:
grafik

Re #38 and #51 (does not fix all cases)


src/person.cpp
```diff
@@ -408,6 +425,11 @@ static uint16 EncodeWalk(const WalkInformation *wi)
 			return (2 << 12) | i;
 		}
 	}
+	for (int i = 0; i < 4; i++) {
+		if (wi == _handyman_water[i]) {
+			return (2 << 12) | (1 << 8) | i;
```
Alberth289346: It's probably useful at some point to fold encoding and decoding these values into some functions, so you have a name what you're doing here, and a name what information you extract.

This also helps if you ever want to change the layout of these values, as finding all literal operations using magic numbers is a nightmare to say the least.

While thinking about it, you could fold the value in a struct, like
``` cpp
struct Thing {
   private int value;
   public setValue(int field1, int field2, ..) { .. }  // Constructor would work too of course.
   public getField1() { return ...; }
   ...
};
```
This should be as large as the number itself, since there are no `virtual` methods. It's also type-safe, you cannot inject a different value in the decoding function.

src/scenery.cpp
```diff
@@ -244,19 +246,28 @@ void SceneryInstance::OnAnimate(const int delay)
...
 }
 
 /**
  * Whether this item is dried up for lack of watering.
  * @return The item is dry.
  */
-bool SceneryInstance::NeedsWatering() const
+bool SceneryInstance::IsDry() const
 {
 	return this->type->watering_interval > 0 && this->last_watered > this->type->watering_interval;
```
Alberth289346: Minor thing: Seem to have missed this before, but comparing `last_watered` against an interval looks odd at first sight, since `last_watered` doesn't express time very clearly (it could be a boolean whether this instance got watered last).

Perhaps `time_since_watered`?  (`last` doesn't look very important to me as code generally doesn't track history, it's also somewhat implied with `since`.)

Alberth289346:
Looks good, some minor things only. Not sure you want to fix these here.

PR #196, opened 2021-05-03
Path Objects

Noordfrees:
This implements path objects: Benches, bins, lamps, litter, and vomit.

grafik

Guests occasionally generate litter and vomit, and handymen clean up after them. Guests will complain when there are dirty paths, demolished objects, and also for a couple of other reasons.

There's still a lot left to do in this area, such as a GUI to actually buy path objects, allowing guests to interact with benches and bins, letting handymen empty the bins, unhappy guests demolishing things, and more; but I think this branch is getting large enough already so all this will have to wait for the next branch.

I also included a commit that refactors the toolbar to use pictorial buttons. It is already wider than the game window at low resolution and the soon-to-come path objects gui button wouldn't fit otherwise.


src/people.cpp
```diff
@@ -95,9 +114,20 @@ void Guests::Uninitialize()
...
+	this->complaint_counter_hunger       = 0;
+	this->complaint_counter_thirst       = 0;
+	this->complaint_counter_waste        = 0;
+	this->complaint_counter_litter       = 0;
+	this->complaint_counter_vandalism    = 0;
+	this->time_since_complaint_hunger    = COMPLAINT_TIMEOUT;
+	this->time_since_complaint_thirst    = COMPLAINT_TIMEOUT;
+	this->time_since_complaint_waste     = COMPLAINT_TIMEOUT;
+	this->time_since_complaint_litter    = COMPLAINT_TIMEOUT;
+	this->time_since_complaint_vandalism = COMPLAINT_TIMEOUT;
```
Alberth289346: You added complaints about 5 different topics, and it looks like you mostly copied 5 times the same code modulo the unique variable names and constants. Did you consider making a `ComplaintTracking` class for a generic complaint, and instantiating that 5 times, possibly even as an array?

Don't quite like my name, maybe you can think of a better one.

src/people.cpp
```diff
@@ -110,11 +140,26 @@ void Guests::Load(Loader &ldr)
...
+				this->complaint_counter_thirst       = ldr.GetWord();
+				this->complaint_counter_waste        = ldr.GetWord();
+				this->complaint_counter_litter       = ldr.GetWord();
+				this->complaint_counter_vandalism    = ldr.GetWord();
+				this->time_since_complaint_hunger    = ldr.GetLong();
+				this->time_since_complaint_thirst    = ldr.GetLong();
+				this->time_since_complaint_waste     = ldr.GetLong();
+				this->time_since_complaint_litter    = ldr.GetLong();
+				this->time_since_complaint_vandalism = ldr.GetLong();
+			}
```
Alberth289346, line 153: This won't get better as you do need to write out all data one way or another, but that's ok.

src/people.cpp
```diff
@@ -280,6 +350,61 @@ void Guests::NotifyRideDeletion(const RideInstance *ri) {
...
+/** A guest complains that many path objects are demolished. May send a message to the player. */
+void Guests::ComplainVandalism()
+{
+	this->complaint_counter_vandalism++;
+	if (this->time_since_complaint_vandalism > COMPLAINT_TIMEOUT && this->complaint_counter_vandalism >= COMPLAINT_THRESHOLD_VANDALISM) {
+		this->complaint_counter_vandalism = 0;
+		this->time_since_complaint_vandalism = 0;
+		_inbox.SendMessage(new Message(GUI_MESSAGE_COMPLAIN_VANDALISM));
+	}
+}
```
Alberth289346, line 346: When you write code like this and use copy/paste (or are tempted to), stop coding, and think if there is a better solution to avoid having X times the same function in a program. It's almost never a good solution.

src/sprite_store.h
```diff
@@ -516,11 +521,25 @@ struct GuiSprites {
...
+	ImageData *toolbar_speed;               ///< Toolbar button speed menu gui sprite.
+	ImageData *toolbar_path;                ///< Toolbar button path build gui sprite.
+	ImageData *toolbar_ride;                ///< Toolbar button ride select gui sprite.
+	ImageData *toolbar_fence;               ///< Toolbar button fence build gui sprite.
+	ImageData *toolbar_scenery;             ///< Toolbar button scenery build gui sprite.
+	ImageData *toolbar_terrain;             ///< Toolbar button terraform gui sprite.
+	ImageData *toolbar_staff;               ///< Toolbar button staff gui sprite.
+	ImageData *toolbar_inbox;               ///< Toolbar button inbox gui sprite.
+	ImageData *toolbar_finances;            ///< Toolbar button finances gui sprite.
+	ImageData *toolbar_objects;             ///< Toolbar button path objects gui sprite.
```
Alberth289346: Did you consider making an array of these sprites? I would think you get the same order of these things everywhere, so you may be able to exploit that in a nice `for` loop at some spots in the code.

Alberth289346: or a computed index, of course.

src/people.cpp
```diff
@@ -23,12 +23,28 @@
...
+		25,  ///< After how many litter    complaints a notification is sent.
+		15,  ///< After how many vandalism complaints a notification is sent.
+};
+static const StringID COMPLAINT_MESSAGES[Guests::COMPLAINT_COUNT] = {  // Indexed by %Guests::ComplaintType.
+		GUI_MESSAGE_COMPLAIN_HUNGRY,     ///< Message for complaints about lack of food.
+		GUI_MESSAGE_COMPLAIN_THIRSTY,    ///< Message for complaints about lack of drink.
+		GUI_MESSAGE_COMPLAIN_TOILET,     ///< Message for complaints about lack of toilets.
+		GUI_MESSAGE_COMPLAIN_LITTER,     ///< Message for complaints about dirty paths.
+		GUI_MESSAGE_COMPLAIN_VANDALISM,  ///< Message for complaints about demolished objects.
+};
```
Alberth289346, line 40: By specifying the size of the array, the compiler will add null entries to extend the array if needed by the enumeration, and you're not warned about it. As those entries probably won't do much useful, you can create bugs by adding new complaints but forgetting to add a line here.

A tighter check can be done with `assert_compile()`, which can be used outside functions, and simply refuses to compile if the assertion fails.

If I remember correctly, if you remove the `Guests::COMPLAINT_COUNT` text, and thus give the compiler room to only make space for the entries that are listed, and add below the closing curly brace add a line `assert_compile(lengthof(COMPLAINT_MESSAGES) == Guests::COMPLAINT_COUNT);`
then the number of entries in the array must match the value of the enumeration or it won't compile.

A simple test is to just comment out one of the lines in the array and trying to compile the code.
I haven't tried it (don't think my dev environment works any more), but it should work.

One small note about `lengthof`, it needs a fully defined array as parameter (that is, not a declaration without having seen a proper definition), or you get nonsense values.

src/people.cpp
```diff
@@ -350,58 +326,20 @@ void Guests::NotifyRideDeletion(const RideInstance *ri) {
...
-		this->complaint_counter_vandalism = 0;
-		this->time_since_complaint_vandalism = 0;
-		_inbox.SendMessage(new Message(GUI_MESSAGE_COMPLAIN_VANDALISM));
+	assert(type < COMPLAINT_COUNT);
+	Complaint &c = this->complaints[type];
+	c.counter++;
+	if (c.time_since_message > COMPLAINT_TIMEOUT && c.counter >= COMPLAINT_THRESHOLD[type]) {
+		c.counter = 0;
+		c.time_since_message = 0;
+		_inbox.SendMessage(new Message(COMPLAINT_MESSAGES[type]));
```
Alberth289346, line 344: Don't know if you are aware of it, but a struct is quite similar to a class, you can have methods in a struct.

Not really needed to change here unless you like the idea, but maybe helpful for some other case.

Alberth289346:
Looks fine to me, feel free to merge when you're done.

src/people.cpp
Noordfrees: Done

Thanks for the review :)

sgtbigman:
Really impressive stuff you are doing here, @Noordfrees!

Noordfrees:
Thanks :)

PR #197, opened 2021-05-06
Path Objects (Part 2)

Noordfrees:
This implements a GUI to place benches, bins, and lamps, and allows guests to interact with them. Handymen empty full bins, guards prevent nearby guests from demolishing park property.


graphics/rcd/lang/en_GB.txt
```diff
@@ -54,6 +54,8 @@ strings {
 		TOOLBAR_GUI_TOOLTIP_FENCE:        "Build fences in your park";
 		TOOLBAR_GUI_SCENERY:              "Scenery";
 		TOOLBAR_GUI_TOOLTIP_SCENERY:      "Build scenery items in your park";
+		TOOLBAR_GUI_PATH_OBJECTS:         "Path Objects";
+		TOOLBAR_GUI_TOOLTIP_PATH_OBJECTS: "Build path objects in your parl";
```
Alberth289346:
```suggestion
		TOOLBAR_GUI_TOOLTIP_PATH_OBJECTS: "Build path objects in your park";
```

src/path.cpp
```diff
@@ -491,7 +492,9 @@ uint8 AddRemovePathEdges(const XYZPoint16 &voxel_pos, uint8 slope, uint8 dirs, P
 						MarkVoxelDirty(ngb_pos[edge]);
 					}
 				}
+				if (PathObjectInstance *obj = _scenery.GetPathObject(ngb_pos[edge])) obj->RecomputeExistenceState();
```
Alberth289346: Don't hide assignments in conditions.

Alberth289346: 3 more below

src/path_object_gui.cpp
```diff
@@ -0,0 +1,203 @@
...
+	void OnClick(WidgetNumber wid, const Point16 &pos) override;
+	void SetWidgetStringParameters(WidgetNumber wid_num) const override;
+
+	void SelectorMouseMoveEvent(Viewport *vp, const Point16 &pos) override;
+	void SelectorMouseButtonEvent(uint8 state) override;
+
+	void SetType(const PathObjectType *t);
+
+private:
+	RideMouseMode path_object_sel;               ///< Mouse selector for building path objects.
```
Alberth289346: `RideMouseMode` ?  Probably functional, but name is weird in this context.

src/path_object_gui.cpp
```diff
@@ -0,0 +1,203 @@
...
+ * Sets what kind of path object type is currently being placed.
+ * @param t Type to select.
+ */
+void PathObjectGui::SetType(const PathObjectType *t)
+{
+	if (this->type == t) {
+		this->type = nullptr;
+	} else {
+		this->type = t;
+	}
```
Alberth289346: Function doesn't do what the description claims, it may end up without selection after a call.

Alberth289346: Could be a one-liner, although this is fine too.

src/path_object_gui.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+	this->GetWidget<LeafWidget>(POBJ_BUTTON_BENCH)->SetPressed(this->type == &PathObjectType::BENCH);
+	this->GetWidget<LeafWidget>(POBJ_BUTTON_BIN  )->SetPressed(this->type == &PathObjectType::LITTERBIN);
+	this->GetWidget<LeafWidget>(POBJ_BUTTON_LAMP )->SetPressed(this->type == &PathObjectType::LAMP);
+
+	if (this->type == nullptr) {
+		this->SetSelector(nullptr);
+	} else {
+		SetSelector(&this->path_object_sel);
```
Alberth289346: One time with `this->` and one time without. Latter looks wrong.

src/person.cpp
```diff
@@ -1617,6 +1664,73 @@ AnimateResult Guest::VisitRideOnAnimate(RideInstance *ri, TileEdge exit_edge)
 	return OAR_CONTINUE;
 }
 
+/** Pixel positions of guests sitting on a bench. */
+static XYZPoint16 _bench_pix_pos[4 /* TileEdge */][2 /* Left = 0, Right = 1 */] = {
+	{XYZPoint16(  0, 160, 0), XYZPoint16(  0,  96, 0)},
+	{XYZPoint16(160, 255, 0), XYZPoint16( 96, 255, 0)},
+	{XYZPoint16(255,  96, 0), XYZPoint16(255, 160, 0)},
+	{XYZPoint16( 96,   0, 0), XYZPoint16(160,   0, 0)},
+};
```
Alberth289346, line 1676: Is this universal, or just happens to be fit for your benches?
If the latter, please consider moving it to the rcd files at some point.

src/person.cpp
```diff
@@ -1617,6 +1664,73 @@ AnimateResult Guest::VisitRideOnAnimate(RideInstance *ri, TileEdge exit_edge)
...
+						if (dynamic_cast<Guard*>(o) != nullptr) {
+							/* A security guard is nearby, so no vandalism just yet. */
+							return OAR_CONTINUE;
+						}
+					}
+				}
+			}
+		}
+
+		obj->Demolish(edge);
```
Alberth289346, line 1719: You ever saw this happen in-game in the original? I never did :)

src/person.cpp
```diff
@@ -1617,6 +1664,73 @@ AnimateResult Guest::VisitRideOnAnimate(RideInstance *ri, TileEdge exit_edge)
...
+	assert(this->activity == GA_RESTING);
+
+	if (this->food > 0 || this->drink > 0 || this->rnd.Uniform(255) > this->nausea) {
+		/* Remain sitting while eating, drinking, or nauseous. */
+		this->StartAnimation(_guest_bench[this->GetCurrentEdge()]);
+		return OAR_OK;
+	}
+
+	/* Get up and keep walking. */
+	this->activity == GA_WANDER;
```
Alberth289346: Nice github action

src/person.cpp
Noordfrees: This is pretty much the only configuration that allows two guests to sit on the same tile edge without overlapping with each other or any neighbouring people, edges or voxels. So I'd consider it universal.

src/person.cpp
Noordfrees: I was never able to actually *see* a guest doing it, so I have no idea if there's an animation for it (I didn't add one). But my benches are forever getting smashed up and then the vandals complain about the extent of vandalism. When I have achieved all objectives and have to spend the remaining year or so sitting around waiting, my main occupation is to go round the park and replace damaged park property.

---

Since we're on the topic, what would you think about enabling mechanics to repair broken path objects? Would be much more convenient than having to seek and replace them all by hand…

src/person.cpp
Alberth289346: You didn't see it?

Screenshot from 2021-05-06 18-54-37

GH action warns that an equality test doesn't do much :)

src/person.cpp
Alberth289346: I never had much trouble with vandalism, only at times when I couldn't buy new rides etc to entertain everybody, or enough handyman to keep the park nice and tidy. I also had low costs for using a ride, so lots of happy customers, I guess :)
Likely my profit was much lower than it needed to be, but who cares. At some point money just pours in by itself.

As for keeping handyman busy, perhaps add a cheats corner, or "enhanced gameplay", or so? Eventually we'll need it anyway :)

src/person.cpp
Noordfrees: Argh, I had overlooked that, and my compiler didn't bother to tell me

PR #199, opened 2021-05-07
Expel guests from benches upon deletion or vandalism

Noordfrees:


Alberth289346:
Didn't know about heap use after free, good thing you did :)

PR #204, opened 2021-06-24
Allow people to walk on pathless land

### LINKED ISSUE #38: Crash when removing a path below a guest

### LINKED ISSUE #51: Ability for guests to walk on some, but not all, pathless lands

Noordfrees:
Finally, here's the result of what I've been working on for over a month. All people can now walk on all types of pathless land, and you can build/remove paths and terraform the land under their feet without them complaining. They will walk around at random (but never climb up or fall down cliffs) and get back on a path if they happen to find one.

Fixes #38 
Fixes #51

### DIFF
--- a/src/person.cpp
+++ b/src/person.cpp
@@ -1690,12 +1690,8 @@ AnimateResult Person::OnAnimate(int delay)
  * @param dz The Z coordinate change between the initial and the current position.
  * @pre The person is located in (but not yet added to) the destination voxel, or somewhere below or above it.
  */
-void Person::DecideMoveDirectionOnPathlessLand(Voxel *former_voxel,
-                                               const XYZPoint16 &former_vox_pos,
-                                               const TileEdge exit_edge,
-                                               const int dx,
-                                               const int dy,
-                                               const int dz)
+void Person::DecideMoveDirectionOnPathlessLand(Voxel *former_voxel, const XYZPoint16 &former_vox_pos,
+		const TileEdge exit_edge, const int dx, const int dy, const int dz)
 {
 	const XYZPoint16 init_pos = this->vox_pos;
 	Voxel *new_voxel = _world.GetCreateVoxel(this->vox_pos, true);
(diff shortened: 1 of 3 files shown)


src/person.cpp
```diff
@@ -1660,7 +1675,89 @@ AnimateResult Person::OnAnimate(int delay)
...
+	/* Convert the bottom part of steep slopes to a format that is easier to handle here. */
+	auto convert_slope = [](TileSlope &slope) {
+		if ((slope & TSB_STEEP) == 0 || (slope & TSB_TOP) != 0) return;
+		const uint8 raised_corner_bit = (slope & ~TSB_STEEP);
+		slope = SL_FLAT;
+		if (raised_corner_bit != TSB_NORTH) slope |= TSB_SOUTH;
+		if (raised_corner_bit != TSB_SOUTH) slope |= TSB_NORTH;
+		if (raised_corner_bit != TSB_WEST ) slope |= TSB_EAST;
+		if (raised_corner_bit != TSB_EAST ) slope |= TSB_WEST;
+	};
```
Alberth289346, line 1710: Complicated calculations!

I think the 1715-1718 lines can be replaced by (modulo renamed variables which you may or may not want to use):
`raised_bottom_bits = (TSB_NORTH | TSB_EAST | TSB_SOUTH | TSB_WEST) ^ raised_top_bit;`
(set all bits, then clear the top-bit).

Bit twiddling is an art :)

src/person.cpp
```diff
@@ -1660,7 +1675,89 @@ AnimateResult Person::OnAnimate(int delay)
...
+			left_height_2  =  this->vox_pos.z + ((new_voxel_slope & TSB_EAST ) == 0 ? 0 : 1);
+			right_height_2 =  this->vox_pos.z + ((new_voxel_slope & TSB_NORTH) == 0 ? 0 : 1);
+			break;
+		case EDGE_NW:
+			left_height_1  = former_vox_pos.z + ((old_voxel_slope & TSB_WEST ) == 0 ? 0 : 1);
+			right_height_1 = former_vox_pos.z + ((old_voxel_slope & TSB_NORTH) == 0 ? 0 : 1);
+			left_height_2  =  this->vox_pos.z + ((new_voxel_slope & TSB_SOUTH) == 0 ? 0 : 1);
+			right_height_2 =  this->vox_pos.z + ((new_voxel_slope & TSB_EAST ) == 0 ? 0 : 1);
+			break;
+		default: NOT_REACHED();
```
Alberth289346: If you first rotate the tile corners so the same edge is always used, these lines collapse in 1/4 of what you have now. `ROR` with 3 parameters is likely the simplest, it allows you to specify number of rotated bits. (Conceptually you rotate counter-clockwise, towards `EDGE_N` for the former voxel.)

I think the current voxel then also rotates counter-clockwise, so `ROR` would work as well.

As for adding, instead of testing for `0` and then adding 0 or 1, you can add the bit you test for after shifting to bit 0, eg
``` cpp
left_height_1  = former_vox_pos.z + ((old_voxel_slope >> TS_NORTH) & 1);  // Doesn't shift, as N is already bit 0.
right_height_1 = former_vox_pos.z + ((old_voxel_slope >> TS_EAST)  & 1);
```
I feel more is possible here, but not sure it's worth while (that may already hold for the test above, perhaps the compiler understand you test a bit and then add 1, it may do the conversion by itself already).

src/person.h
```diff
@@ -157,6 +157,12 @@ class Person : public VoxelObject {
 	uint8 GetInparkDirections();
 
 	virtual void DecideMoveDirection() = 0;
+	void DecideMoveDirectionOnPathlessLand(Voxel *former_voxel,
+	                                       const XYZPoint16 &former_vox_pos,
+	                                       const TileEdge exit_edge,
+	                                       const int dx,
+	                                       const int dy,
+	                                       const int dz);
```
Alberth289346: Don't think this is the code style we aim for, but I can see the use of it.

src/person.cpp
Noordfrees: I reverted the change now as actually this introduced a logical error, because the corner we want to remove from the bitset is the one *opposite* the raised corner (e.g. for a North steep slope the result should be N|W|E).

src/person.cpp
Noordfrees: I changed it to something even shorter (though slightly less readable), how about that

src/person.h
Noordfrees: I had initially made this function `virtual` which made the line unreasonably long, but now I collapsed it into the usual style again.

src/person.cpp
Alberth289346: Assuming you mean "S|W|E":

(all bits set xor the top) means all bits set except the top-bit  (1 xor 1 = 0)

Alberth289346: oh, I see

`TSB_MASK_ALL ^ ROR(raised_corner_bit, 4, 2);` or `ROL` of course, 2 shifts in a 4 bit number works in either direction.

src/person.cpp
```diff
@@ -1711,40 +1707,21 @@ void Person::DecideMoveDirectionOnPathlessLand(Voxel *former_voxel,
...
-			left_height_1  = former_vox_pos.z + ((old_voxel_slope & TSB_WEST ) == 0 ? 0 : 1);
-			right_height_1 = former_vox_pos.z + ((old_voxel_slope & TSB_NORTH) == 0 ? 0 : 1);
-			left_height_2  =  this->vox_pos.z + ((new_voxel_slope & TSB_SOUTH) == 0 ? 0 : 1);
-			right_height_2 =  this->vox_pos.z + ((new_voxel_slope & TSB_EAST ) == 0 ? 0 : 1);
-			break;
-		default: NOT_REACHED();
+	uint16 node_heights[4];
+	for (uint8 i = 0; i < 4; i++) {
+		const uint8 bitshift = (exit_edge + i) % 4;
+		node_heights[i] = (i < 2 ? former_vox_pos : this->vox_pos).z + (((i < 2 ? old_voxel_slope : new_voxel_slope) & (1 << bitshift)) >> bitshift);
```
Alberth289346: `(((i < 2 ? old_voxel_slope : new_voxel_slope) & (1 << bitshift)) >> bitshift)` Shifting 1 left, masking and then shifting right again can also be done by shifting right and then masking the lowest bit `(((i < 2 ? old_voxel_slope : new_voxel_slope) >> bitshift) & 1)`

Noordfrees: All implemented :)

Issue #149, opened 2021-01-20
[Discussion] Project Status


KevDoy:
What's the realistic status on FreeRCT? 

I know activity was rather slow back in 2016 when I contributed some pixel art, but I am curious if we'll see the project pick up again in the future or if it's considered dormant (or worse).

Cheers,

Kev

Alberth289346:
Who knows.

After 5 years I concluded I would have to drag the thing forward for another 3-5 years, which was just too much to do by just me.
Clearly something went wrong, if only I knew what.

I do still ponder about alternative paths every now and then, but I have currently no space in my life to actually try it.

Noordfrees:
Hi

A while back I stumbled over FreeRCT and liked the idea a lot, being a fan of the original RCT1. I have experience coding C++ and if someone would point me to something suited for a newcomer, I could try and contribute code to the project :) 

Best regards 
Benedikt

LordAro:
Unfortunately, the project is essentially dead. Everyone involved either got tired of it or ran out of time. You're welcome to work on what's there though, I don't particularly have any issues assigning ownership to others if that's desirable.

The real answer here is that you should go look at https://github.com/OpenRCT2/OpenRCT2. It was first announced about a year after we got going with FreeRCT and it's been extremely successful - it's quite likely hoovered up any chance of other contributors coming back to this.

Noordfrees:
Thanks for the reply. I don't mind contributing to a project with few active developers – maybe I can help revive it :) I think I'll start by investigating the assert fail I just encountered in master, and then work on adding some features.

What I don't like about OpenRCT2 is that they provide only the engine and require the original game files. A truly free game should be modder-friendly and never depend on proprietary assets. Even if OpenRCT2 is bigger and more successful, a fully free and independent new-implementation like FreeRCT seems way more appealing to me.

LordAro:
I imagine they'll get there eventually. OpenTTD (which has been made in much the same way) was dependent on the original TTD files for a quite some time as well. (Well, 6 years, when it's now nearly 17 years old, but still)

Newbytee:
> What I don't like about OpenRCT2 is that they provide only the engine and require the original game files. A truly free game should be modder-friendly and never depend on proprietary assets

There's this: https://github.com/OpenRCT2/OpenGraphics Not sure how complete it is at this point.

Alberth289346:
Some thoughts what could be done differently.

[...]

PR #207, opened 2021-07-02
Replace debian building rules with CPack & CI

Noordfrees:
See https://github.com/FreeRCT/FreeRCT/issues/149#issuecomment-872534613

How should we proceed regarding Ubuntu packaging? I'd propose to set up a FreeRCT-dev team on launchpad, and a GitHub action that syncs GitHub master to a Launchpad branch on every new push event and triggers automated daily builds to an official PPA belonging to the FreeRCT team. Opinions?


Alberth289346:
It's likely simpler in the long run to have someone from ubuntu building the release. Then it gets added to their repository and their distribution, in the standard package manager configuration. You then also get all derived distributions as well, for as far as they use the ubuntu repositories.

Having a working build helps perhaps, don't have much experience in these matters.

@LordAro  Suggestions? Ideas?

Alberth289346:
Some advertising may be useful, so people know it exists and isn't dead.
I used to post short blogs, at the time.

Ie I think getting a build is the simple part, getting known as being alive is more complicated.

Don't know if there are additional access rights, but as far as I am concerned, you can do things in name of the project.

README.rst
```diff
@@ -23,20 +23,31 @@ Building is as simple as
 ::
 
         $ git clone https://github.com/FreeRCT/FreeRCT.git
-        $ cd FreeRCT # Go into the downloaded source directory.
-        $ cmake .    # Checks libraries are where they're supposed to be and replaces some strings
-        $ make       # Let make do the heavy work.
+        $ cd FreeRCT    # Go into the downloaded source directory.
+        $ cmake .       # Checks libraries are where they're supposed to be and replaces some strings.
```
Alberth289346: While it's great this works imho, officially cmake only does out-of-source builds.

It is generally expected that something like
``` bash
$ mkdir build
$ cd build
$ cmake ../FreeRCT
$ make
$ make install
```
works.

If it doesn't, then either make it work or warn against it in the documentation and make it fail if tried anyway.

Perhaps advertise this as an extra nice feature :)

README.rst
```diff
@@ -23,20 +23,31 @@ Building is as simple as
...
-OVERRIDE_VCS      OFF     Override the lack of a VCS checkout.
-ASAN              OFF     Use AddressSanitizer (https://clang.llvm.org/docs/AddressSanitizer.html).
-USERDATA_PREFIX   '.'     Directory where user data such as savegames is stored.
-================= ======= ==========================================================================
+======================= ============================= ================================================================================
+Name                    Default value                 Explanation
+======================= ============================= ================================================================================
+OVERRIDE_VCS            OFF                           Override the lack of a VCS checkout. Supported VCSs are git and bzr.
+VERSION_STRING          '0.0.0-norev'                 If OVERRIDE_VCS is enabled, use this string as the build version.
+ASAN                    OFF                           Use AddressSanitizer (https://clang.llvm.org/docs/AddressSanitizer.html).
```
Alberth289346: As I didn't have any idea what ASAN is, this line reads like: `'X': "use X (see http://.../X)"`. This line doesn't explain what X does, or why X is useful, or who is supposed to use this X thing. From only "X: use X", there is no clue for deciding whether I should make the effort of adding this option.  
You provided a full name and link and that's great, but "address" is a quite broad topic, and combined with 'sanitize' I went towards normalizing URLs or email addresses, which doesn't make sense to me at all.  
I can open the web page, but it would be very useful to at least know beforehand what the topic is going to be.

Note that the above is a common trap for people writing documentation,  you see it everywhere if you know it exists. The trap is the implicit assumption that all readers have the same knowledge as the author. Obviously, that cannot be true (fortunately).

----
There are 2 ways out of this. One option is to explain somewhere what ASAN is in a few words. That gives readers a fighting chance of deciding whether ASAN is of use to them.

Another option is to generalize a bit. You don't want `ASAN`, you want a `MEMORYCHECKER` (or `MEMCHECK` or so), and today we support `ASAN`, ie `MEMORYCHECK=ASAN`. That gives much more information to a reader. Even then, likely the goal of 'MEMORYCHECK' needs to be explained as there is room for different interpretations as well. (Checking whether the memory itself is sane versus sane behavior of the program towards memory use.)

README.rst
```diff
@@ -23,20 +23,31 @@ Building is as simple as
...
-================= ======= ==========================================================================
-OVERRIDE_VCS      OFF     Override the lack of a VCS checkout.
-ASAN              OFF     Use AddressSanitizer (https://clang.llvm.org/docs/AddressSanitizer.html).
-USERDATA_PREFIX   '.'     Directory where user data such as savegames is stored.
-================= ======= ==========================================================================
+======================= ============================= ================================================================================
+Name                    Default value                 Explanation
+======================= ============================= ================================================================================
+OVERRIDE_VCS            OFF                           Override the lack of a VCS checkout. Supported VCSs are git and bzr.
+VERSION_STRING          '0.0.0-norev'                 If OVERRIDE_VCS is enabled, use this string as the build version.
```
Alberth289346: Does this mean that `OVERRIDE_VCS` and `VERSION_STRING` always occur as pair? In that case, woud it make sense to combine both to a single option? (`VERSION_STRING` seems the most likely candidate to me, but maybe there are some edge cases I am missing.)

README.rst
```diff
@@ -23,20 +23,31 @@ Building is as simple as
...
+ASAN                    OFF                           Use AddressSanitizer (https://clang.llvm.org/docs/AddressSanitizer.html).
+CMAKE_INSTALL_PREFIX    '/usr'                        Directory where 'make install' installes 
+USERDATA_PREFIX         '~/.freerct' (Windows) or     Directory where user data such as savegames is stored.
+                        '~/.config/freerct' (other)
```
Alberth289346: You may want to make a note below saying that `~` is expanded by FreeRCT itself, to avoid packagers believing the documentation is silently assuming a user-only install.

(Packagers want a runtime expansion of `~` and we provide that, but without telling here what `~` actually does here, any Unix packager will assume the example uses shell expansion at  build time.)

EDIT: That is, they are not going to dig in source to see how it actually works, in general.

README.rst
```diff
@@ -23,20 +23,31 @@ Building is as simple as
...
+======================= ============================= ================================================================================
+
+Note that CMake options are prefixed with '-D', so the full call is e.g.
+
+::
+
+        $ cmake -DASAN=ON -DCMAKE_INSTALL_PREFIX='~/.local' .
```
Alberth289346: +1  !

CMake end-user documentation is beyond horrible.

You may want to extend the above example install sequence with the install prefix, as people typically need that anyway, also giving the default value and again telling what `~` means (as it means something else?)

debian/changelog
```diff
@@ -1,3 +1,9 @@
+freerct (1:0.1~bzr5) UNRELEASED; urgency=medium
+
+  * Made everything work again after years of disuse.
```
Alberth289346: You also added several new things, would be nice if you mentioned those as well imho. (Doesn't need to be complete, but a few highlights, showing progress is being made rather than just update it to current software versions.

debian/freerct.desktop
```diff
@@ -2,9 +2,13 @@
 Type=Application
 Encoding=UTF-8
 Name=FreeRCT
+Name[de]=FreeRCT
 Name[es]=FreeRCT
-Comment=FreeRCT aims to be a free and open source game which captures the look, feel and gameplay of the popular games RollerCoaster Tycoon 1 and 2.
+Name[nds]=FreeRCT
+Comment=FreeRCT aims to be a free and open source game which captures the look, feel and gameplay of the popular games RollerCoaster Tycoon 1
```
Alberth289346: `and 2.` seems to be missing here.

src/CMakeLists.txt
```diff
@@ -128,12 +128,20 @@ IF(GIT_FOUND AND IS_DIRECTORY "${CMAKE_SOURCE_DIR}/.git")
...
 	                WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
 	)
-	string(STRIP "${VERSION_STRING}" VERSION_STRING) # Newline
+	string(STRIP "0.1~git${VERSION_STRING}" VERSION_STRING) # Newline
+ELSEIF(IS_DIRECTORY "${CMAKE_SOURCE_DIR}/.bzr")
+	execute_process(COMMAND bzr revno
+	                OUTPUT_VARIABLE VERSION_STRING
+	                WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
+	)
+	string(STRIP "0.1~brz${VERSION_STRING}" VERSION_STRING) # Newline
```
Alberth289346: Not sure why you added this.

I used `bzr` for a while, but that is ages ago. I am fine if we just drop this (but fine if you want to keep it as well).

I also used `hg` and I think that VCS used to be a lot bigger, yet you decided not to add it (again, I am fine either way, just curious why you made these choices).

src/CMakeLists.txt
```diff
@@ -128,12 +128,20 @@ IF(GIT_FOUND AND IS_DIRECTORY "${CMAKE_SOURCE_DIR}/.git")
...
 ELSE()
 	# Only do this if you really want to
 	IF(OVERRIDE_VCS)
 		message(WARNING "WARNING: No VCS checkout detected but overridden by user")
-		set(VERSION_STRING "0.0.0-norev")
+		if (NOT DEFINED VERSION_STRING)
+			set(VERSION_STRING "0.0.0-norev")
+		ENDIF()
 	ELSE()
 		message(FATAL_ERROR "No VCS checkout detected. Enable the OVERRIDE_VCS option if this is intended. This is not recommended.")
```
Alberth289346: Hmm, yeah.

I am not sure this is such a big thing to abort building. Also, what's the value of a version string if it's trivial to state any version you want anyway?

Older VCSes supported `$Id$` where you could insert a source file version string in the code, but this practice has been abandoned by `git`. As such, finding what was actually used to build a binary is near impossible nowadays.

Saying it in a different way, is it bad if we would just print a warning "he, we're missing a VCS checkout, and use 'unknown' as version", and then just build it?

We'd be missing the users that have a compiler installed and know about cmake and make and stuff, but don't pull the source from a repository. Seems to me that is a very small fraction of users. Sending source code around by mail isn't common any more.

what do you think?

README.rst
Noordfrees: Tested it and it works, so I changed the example accordingly.

README.rst
Noordfrees: Updated the Readme with a section about ASan. It's a very useful development tool so a paragraph on it can't hurt.

README.rst
Noordfrees: I removed OVERRIDE_VCS and changed the priority order. If VERSION_STRING is set, this one is always used; next try to get the git revision; finally fall back to `0.0.0-revdetect-broken`.

README.rst
Noordfrees: done

README.rst
Noordfrees: done

debian/freerct.desktop
Noordfrees: fixed

src/CMakeLists.txt
Noordfrees: Removed it again. For launchpad mirroring bzr is needed, but since the PPA wants to specify a custom version string anyway this was unnecessary.

debian/changelog
Noordfrees: I think the `debian/changelog` only refers to directly packaging-related changes, i.e. changes in the debian/ directory. All other changes should better get a separate ChangeLog file in the parent directory. But that's something for another branch, imho.

Noordfrees:
> It's likely simpler in the long run to have someone from ubuntu building the release. Then it gets added to their repository and their distribution, in the standard package manager configuration. You then also get all derived distributions as well, for as far as they use the ubuntu repositories.
> 
> Having a working build helps perhaps, don't have much experience in these matters.

While we should definitely aim to get FreeRCT into the official debian and ubuntu archives, I think *completely* separating packaging from the main repo would do more harm than good, because then we couldn't run our own PPA. 
Over at Widelands, our stable releases get dozens of thousands of downloads, but nearly all bug reports and feature suggestions come from two dozen or so people who use the development versions. I estimate that about a third of them compiles from source, another third uses the daily PPA, and the others the daily Windows and every-few-days Mac binaries. So seeing that the FreeRCT userbase is rather small :wink: a daily PPA would in my opinion be an important feature for us. I see two ways to achieve it:
1. Keep and maintain the debian/ directory here in the main repo, and set up a Launchpad mirror that copies it and builds the daily PPA from it. That's how we're doing it at Widelands so we can just reuse the integrations code I wrote there.
2. Remove the debian/ directory from the main repo, and keep a separate packaging repo on Launchpad that syncs itself with the main repo every day. Similar workflow though slightly more complex to maintain.

When/if a debian developer starts packaging FreeRCT officially, they'll probably create their own repository with necessary packaging changes, which we'd in turn need to sync back to our debian/ directory (and which is far less work than it sounds like).
I'd be fine with setting up either way, what sounds better?

---

> Some advertising may be useful, so people know it exists and isn't dead.
> I used to post short blogs, at the time.
> 
> Ie I think getting a build is the simple part, getting known as being alive is more complicated.
> 
> Don't know if there are additional access rights, but as far as I am concerned, you can do things in name of the project.

Yes… I'm afraid I don't know any social media/blogs/linux newspapers/whatever so I'm not sure how much I can really do for this…
A real homepage with a community forum would probably be a good way to attract users and get more feedback. As a start how about using [GitHub Pages](https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages) to create some nice screenshots, online manual, etc?

README.rst
```diff
@@ -23,20 +23,50 @@ Building is as simple as
...
 
         $ git clone https://github.com/FreeRCT/FreeRCT.git
-        $ cd FreeRCT # Go into the downloaded source directory.
-        $ cmake .    # Checks libraries are where they're supposed to be and replaces some strings
-        $ make       # Let make do the heavy work.
+        $ cd FreeRCT               # Go into the downloaded source directory.
+        $ mkdir build && cd build  # Create and enter a build directory.
+        $ cmake ..                 # Checks libraries are where they're supposed to be and replaces some strings.
+        $ make                     # Let make do the heavy work.
+        $ make install             # Optional step to install the program on the system. May require root privileges.
```
LordAro, line 31: Can be simplified to 
```suggestion
        $ cmake -B build ..        # Checks libraries are where they're supposed to be and replaces some strings.
        $ make -C build            # Let make do the heavy work.
        $ make -C build install    # Optional step to install the program on the system. May require root privileges.
```

debian/changelog
```diff
@@ -1,3 +1,9 @@
+freerct (1:0.1~bzr5) UNRELEASED; urgency=medium
```
LordAro: bzr? This isn't anything to do with bazaar...

LordAro (commented):
I think as much as a PPA is unnecessary, tbh. A deb is just as easy to install (it's not like there's multiple packages), which can just be published in GitHub releases. That said, daily GH releases would be a pain to maintain (still possible without actually tagging anything, but more irritating for people to update)

Instead of maintaining all the debian/* stuff, OTTD has moved to just using `cpack` to generate everything - should consider doing the same here, rather than faffing around with internal debian stuff that would only get replaced by any official packaging anyway - https://github.com/OpenTTD/OpenTTD/commit/b7643b1d36cb41eb6712042761c13b9819fcbe74

Alberth289346:
> I'd be fine with setting up either way, what sounds better?

I would say keep everything in a single repository, at  least for now. Keeping debian build configuration elsewhere can only complicate matters.

What we actually need, I don't know, LordAro has much better insights in this matter.

> Yes… I'm afraid I don't know any social media/blogs/linux newspapers/whatever

That's easily solved! :p 
https://freerct.blogspot.com/  is what we wrote. As you can see, it's mostly about user-visible changes with a picture and some details. Nothing deep or complicated.

> As a start how about using GitHub Pages to create some nice screenshots, online manual, etc?

As long as the source is in the repository, I am fine with that. The reason for that is that it's much simpler to keep in sync with development. Also you can have a manual for a specific version of the game.

README.rst
```diff
@@ -23,30 +23,49 @@ Building is as simple as
...
-ASAN                    OFF                           Use AddressSanitizer (https://clang.llvm.org/docs/AddressSanitizer.html).
+======================= ============================= ================================================================================================
+VERSION_STRING          -                             Use this string as the build version. If not specified, the version is detected
+                                                      automatically from git. If this also fails, the version defaults to '0.0.0-revdetect-broken'.
+ASAN                    OFF                           Use AddressSanitizer (see below).
 CMAKE_INSTALL_PREFIX    '/usr'                        Directory where 'make install' installes 
 USERDATA_PREFIX         '~/.freerct' (Windows) or     Directory where user data such as savegames is stored.
-                        '~/.config/freerct' (other)
-======================= ============================= ================================================================================
+                        '~/.config/freerct' (other)   Note that in the USERDATA_PREFIX variable expansion of the '~' character to the
```
Alberth289346:
```suggestion
                        '~/.config/freerct' (other)   Note that '~' character expansion in the USERDATA_PREFIX variable to the
```

Alberth289346:
Stuff looks good, please consider the suggested change, and feel free to merge if we're done discussing. Alternatively, we could move the discussion elsewhere.

Noordfrees:
This has lain around for a while now…
I did look at cpack, and IMHO the one disadvantage compared to a PPA is publishing the releases. We have three options, as far as I see:

1. Releasing only "stable" builds (0.1, we're getting closer…); anyone who wants a snapshot of the latest development version has to compile from source (easiest but user-hostile).
2. Providing daily builds over a GitHub action. Actions can upload artifacts which are stored for public download for 90 days ([example](https://github.com/widelands/widelands/actions/runs/1128467535#artifacts)). Bonus is, this way we can publish not only standalone debian/ubuntu packages but also daily-build installers for Win and Mac as well (provided someone tests them).
3. A daily PPA, so we need to maintain the debian build stuff here.

The suggestion of @LordAro goes in the direction of 2), and if we decide to set up packaging workflows for all mentioned platforms this might be easier to maintain as well if we don't use Launchpad. On the other hand, users will then need to update their daily builds manually every couple of days (assuming a speed of development higher than last month ;) ), whereas a PPA would update the users' installation automatically with every routine update.

So, should I drop all Launchpad/PPA integration and move to a GitHub Action with CPack, or keep it as it's now and add LP mirroring?

Noordfrees:
Btw I also started some work on a GitHub pages site in my fork. Still very incomplete so far. https://noordfrees.github.io/FreeRCT/

Noordfrees:
I just implemented the CPack approach now. It works for Linux (deb and tarball, tested) and Windows (exe and zip, UNTESTED); couldn't get it to work on MacOS yet. The packages are uploaded automatically by the GitHub action CI [like this](https://github.com/FreeRCT/FreeRCT/actions/runs/1139564230#artifacts).

Noordfrees:
And I added another fix in 2df22de80f355dcb767b49064e45149298bb1ca6 that needs review

Alberth289346:
I think @LordAro  would be the best person to have a look, unfortunately, he seems to be on holiday currently.

Noordfrees:
This PR has now been lying around for half a year since the last change. I think that this and #225 are the only two things that need to be merged before we could have a first 0.1 release.

Alberth289346:
Let's just merge it, mr Aro isn't likely to appear apparently.

LordAro:
M'bad. Looks fine.

`/debian` gitignores could've just been a single line (it acts recursively), but not an issue

PR #208, opened 2021-08-19
Fix language-specific rides saveloading

Noordfrees:
Nobody except me has tested this game for at least 4½ months, otherwise someone would have noticed that FreeRCT crashes on startup if your system language is not set to low german :P 

*Wenn man nich doch elkeen enkelt vun jo all Plattdüütsch snackt; dann nehm ick all torügg ;)*

There were two bugs here: Using the localized!!! ride type name to persist ride instances, and dereferencing a `nullptr` if this lookup failed.


Alberth289346:
We can definitely use more tests :)

PR #210, opened 2021-08-24
New Website

### LINKED ISSUE #209: New Website
I've written a simple website for FreeRCT using GitHub Pages. The result is at https://noordfrees.github.io/FreeRCT/. Play around a bit on the various subpages and try it at different browser window sizes :)

Of course there's still room for improvement. Also some parts of the stuff in `doc/` are now superfluous and should probably be moved to the website folder completely.

The code is in https://github.com/Noordfrees/FreeRCT/tree/gh-pages/docs.

- Do we want to move the site from my fork to the main repo and make it official?
- Feedback, opinions, suggestions appreciated :)

Noordfrees:
closes #209


Alberth289346:
Oh you're so quick :)

I had another look, and was planning to make comments about that, but something else intervened :) 

- You may want to have another look at the menu bar at the home-page.
  People generally start reading at the left and stop reading at the right, so the important bits you want to get noticed should be "left"-ish, and the less relevant things "right"-ish. Currently, you seem to have ordered on when people need information, and that makes sense. However, is "news archive" more needed than "contribute"? I think at least those two should be swapped (you want people to read "contribute" before reading ancient news, I think).

  Then it becomes a matter of priority. Is "screen shots" more important than "contribute"? Difficult question I found. Alternatively, you could move "screen shots" before "get it!", people like looking at pictures (especially pictures of cats, but we don't have those). A few more teasers before installing it :) 

- I read the "Testing" part wrong in the contributions and got confused. The "You can also test open PRs ..." sentence is a new topic, and should start at a new paragraph (I read it as belonging to testing the program and then got confused about having to use a PR for reporting about the test).

  You're likely over-estimating the capabilities of testers. In my experience, compiling code is generally beyond them. If they can do that, they are likely coders already, but until we build binaries for these, it's the only option we can ask.

  There is one other thing they can do, which is triaging bug reports. Often you get bug reports that have no nice way to reproduce, finding a reproduction sequence is generally not hard, just time consuming. So you suggest that here as well.

- As a check, is the `~` syntax correct in the "compile from source" description for building? I just typed that, but not sure if it actually works in this way as intended, or does the `~` need more escaping? (the command is read by the shell you type, cmake gets it, puts it in a makefile, those commands are executed by a shell from make, all steps where the `~` may disappear).

Otherwise, all changes look fine to me, feel free to merge.

Issue #212, opened 2021-08-26
Code Modernizations


Noordfrees:
There are a few techniques in the codebase which strike me as odd. I'd be interested to hear if there's a reason why it's done this way and if it wouldn't be better to "modernize" them.

### Fixed-size arrays:
https://github.com/FreeRCT/FreeRCT/blob/b05aa4ed20dab086b970019561c1d190bcdc1892/src/ride_type.h#L282-L285
This just places completely unnecessary arbitrary limitations on park size. In my opinion, all such arrays should be replaced with `std::vector` or – for `RideInstance`s – `std::map`.

### Strings:
All our strings are `uint8*` currently. Why?? What is the reason for not using the ordinary C-style `char*`? Then we wouldn't need to `reinterpret_cast` or even C-style-cast all the time whenever we access library functions.
https://github.com/FreeRCT/FreeRCT/blob/b05aa4ed20dab086b970019561c1d190bcdc1892/src/widget.cpp#L673-L675
Or even better, migrate to `std::string`, which has much better high-level handling, especially for lookup and modification.

### Raw pointers:
Are prone to memory leaks, double-frees, and alloc-dealloc-mismatches. In my opinion, the code should contain as few `delete` statements as possible, and memory ownership should always be handled by `std::unique_ptr`, or where absolutely necessary `std::shared_ptr`.
https://github.com/FreeRCT/FreeRCT/blob/b05aa4ed20dab086b970019561c1d190bcdc1892/src/track_piece.cpp#L172-L178

Alberth289346:
C++ has moved a lot since I started.

At the time, C++11 existed, I was not familiar with it, and I had my hands full without having to learn the language as well. Therefore I used the style of OpenTTD at the time, "old" C++ mostly (I think).

Today, I would indeed use a more modern C++ version for coding.

As for your points, we seem to agree with a few notes.

### Fixed size arrays

I agree that lots of those `[]` variables can be replaced by a vector, basically anything where you don't know or care how many elements you have. I am not sure it holds for everything, Eg a new type of ride isn't trivial to add, so I see no value in making that dynamic, and it may even be slower, since vector allocates the actual data elsewhere, so you get an additional memory dereference with associated cache misses.

So yes, but maybe not everything.

### Strings

Signedness of `char` is implementation dependent. You only get guarantees for the `0--127` range, which isn't enough. So `uint8` was used to resolve that problem.
[...]

PR #213, opened 2021-08-27
Convert some fixed-size arrays to vectors or maps

Noordfrees:
Re #212 

This also fixes a bug with saveloading rides-related data after removing ride instances: Saveloading code assumes that every ride instance has a unique index that never changes, which was not guaranteed until now.

### DIFF
--- a/src/coaster_gui.cpp
+++ b/src/coaster_gui.cpp
@@ -524,8 +524,8 @@ void CoasterInstanceWindow::OnClick(WidgetNumber widget, const Point16 &pos)
 
 		case CIW_CHOOSE_ENTRANCE: {
 			DropdownList itemlist;
-			for (int i = 0; _rides_manager.entrances[i] != nullptr; i++) {
-				_str_params.SetUint8(1, _language.GetText(_rides_manager.entrances[i]->name));
+			for (auto &eetype : _rides_manager.entrances) {
+				_str_params.SetUint8(1, _language.GetText(eetype->name));
 				itemlist.push_back(DropdownItem(STR_ARG1));
 			}
 			this->ShowDropdownMenu(widget, itemlist, this->ci->entrance_type, COL_RANGE_DARK_RED);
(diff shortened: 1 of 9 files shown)


src/coaster_gui.cpp
```diff
@@ -524,17 +524,17 @@ void CoasterInstanceWindow::OnClick(WidgetNumber widget, const Point16 &pos)
 
 		case CIW_CHOOSE_ENTRANCE: {
 			DropdownList itemlist;
-			for (int i = 0; _rides_manager.entrances[i] != nullptr; i++) {
-				_str_params.SetUint8(1, _language.GetText(_rides_manager.entrances[i]->name));
+			for (auto &eetype : _rides_manager.entrances) {
```
Alberth289346: We got the option to use `const` data here, is that something to aim for? (I'd say yes, but maybe you have other plans?)

src/ride_type.cpp
```diff
@@ -681,18 +681,10 @@ void RideInstance::Save(Saver &svr)
...
-	std::fill_n(this->exits, lengthof(this->exits), nullptr);
 }
 
 RidesManager::~RidesManager()
 {
-	for (uint i = 0; i < lengthof(this->entrances); i++) delete this->entrances[i];
-	for (uint i = 0; i < lengthof(this->exits); i++) delete this->exits[i];
-	for (uint i = 0; i < lengthof(this->ride_types); i++) delete this->ride_types[i];
-	for (uint i = 0; i < lengthof(this->instances); i++) delete this->instances[i];
 }
```
Alberth289346: Constructor can be dropped. Destructor can be dropped too, it doesn't look like a baseclass at least.

src/coaster_gui.cpp
Alberth289346: You seem to use `const` at other places, so this one is likely a bad suggestion.

Noordfrees: As a rule I try to use `const` wherever possible – it's a good way to discover bugs already at compile-time. Only variables that are actually meant to be modified should be non-`const`.

PR #214, opened 2021-08-29
Replace some raw deletes with `unique_ptr`

Noordfrees:
Re #212

### DIFF
--- a/src/map.h
+++ b/src/map.h
@@ -64,6 +64,12 @@ public:
 	uint8 instance;       ///< Ride instances that uses this voxel.
 	uint16 instance_data; ///< %Voxel data of the #instance stored here.
 
+	/** Constructor */
+	explicit Voxel()
+	{
+		this->ClearVoxel();
+	}
+
 	/**
 	 * Get the ride instance at this voxel.
 	 * @return Ride instance at this voxel.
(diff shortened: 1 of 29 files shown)


src/map.h
```diff
@@ -538,7 +537,7 @@ class VoxelStack {
 	void Save(Saver &svr) const;
 	void Load(Loader &ldr);
 
-	Voxel *voxels;   ///< %Voxel array at this stack.
+	std::unique_ptr<Voxel[]> voxels;  ///< %Voxel array at this stack.
```
Alberth289346: Wouldn't a vector be more useful here, or is that a next step?

src/sprite_store.h
```diff
@@ -32,25 +32,20 @@ class ImageData;
...
+	virtual ~RcdBlock() = default;
 };
 
 typedef std::map<uint32, ImageData *> ImageMap; ///< Map of loaded image data blocks.
 
 /** Texts of objects. */
 class TextData : public RcdBlock {
 public:
-	TextData();
-	~TextData();
```
Alberth289346, line 47: I am wondering whether you should make all destructors just `= default`. Might be a bit more work and not always needed, but at least it is then never missing and never forgotten.

Alberth289346:
Sorry for not finding this earlier, I don't visit GH much these days.

Don't think I am in a position to properly verify what you did, but I assume you checked with a compiler and memory checker.

The only things I wonder about somewhat are lack of vectors (but that may be a next step to you), and whether simply have virtual destructors everywhere makes sense.

Otherwise seems fine.
Not sure if you want to fix anything wrt to my remarks, if not, please merge.

src/map.h
Noordfrees: Changed. This also allowed me to rip out a lot of copy code that duplicated and replaced the whole stack every time a new voxel was added to the top or bottom; now we can just construct the new element in-place.

src/sprite_store.h
Noordfrees: Nothing to gain, IMHO. Not declaring a custom ctor or dtor is the same as declaring it as `default`. Only base classes need a dtor that is explicitly marked `virtual`, for other classes it can be omitted to remove clutter.

PR #215, opened 2021-09-05
Refactor `uint8*` to `std::string` or `char*`

Noordfrees:
Re #212
This refactors most (but not all) uses of `uint8*` to `std::string` or `char*`. Actually some (most) parts of our string magic that now use `char*` could do with a full rewrite from scratch using `std::string`, but that's out of scope for this branch.

### DIFF
--- a/src/config_reader.cpp
+++ b/src/config_reader.cpp
@@ -80,6 +80,16 @@ ConfigSection::~ConfigSection() {
 }
 
 /**
+ * Check if this section contains an element with the specified key.
+ * @param key Value of the key to look for (case sensitive).
+ * @return An item with this key exists.
+ */
+bool ConfigSection::HasItem(const std::string &key) const
+{
+	return this->items.count(key) > 0;
+}
+
+/**
  * Get an item from a section if it exists.
  * @param key Value of the key to look for (case sensitive).
  * @return The associated item if it exists, else \c nullptr.
@@ -142,8 +152,9 @@ ConfigFile::ConfigFile(const std::string &fname) : filename(fname)
 			if (*line2 == ']') {
 				*line2 = '\0';
 				const std::string sect_name(StripWhitespace(line + 1, line2));
-				if (this->sections.count(sect_name) == 0) this->sections.emplace(sect_name,
-						std::unique_ptr<ConfigSection>(new ConfigSection(*this, sect_name)));
+				if (this->sections.count(sect_name) == 0) {
+					this->sections.emplace(sect_name, std::unique_ptr<ConfigSection>(new ConfigSection(*this, sect_name)));
+				}
 				current_sect = this->sections.at(sect_name).get();
 			}
 			continue;
@@ -169,6 +180,18 @@ ConfigFile::ConfigFile(const std::string &fname) : filename(fname)
 }
 
 /**
+ * Check if a section in this config file contains an element with the specified key.
+ * @param sect_name The section in which to look for the key (case sensitive).
+ * @param key Value of the key to look for (case sensitive).
+ * @return An item with this key exists in this section.
+ */
+bool ConfigFile::HasValue(const std::string &sect_name, const std::string &key) const
+{
+	ConfigSection *s = this->GetSection(sect_name);
+	return s != nullptr && s->HasItem(key);
+}
+
+/**
  * Get a section from the configuration file.
  * @param sect_name Name of the section (case sensitive).
  * @return The section with the give name if it exists, else \c nullptr.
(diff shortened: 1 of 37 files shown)


src/config_reader.cpp
```diff
@@ -125,63 +108,43 @@ static char *StripWhitespace(char *first, char *last = nullptr)
...
 			char *line2 = line;
 			while (*line2 != '\0' && *line2 != ']') line2++;
 			if (*line2 == ']') {
 				*line2 = '\0';
-				/* XXX Look for an existing config section? */
-				current_sect = new ConfigSection(StripWhitespace(line + 1, line2));
-				this->sections.push_front(current_sect);
+				const std::string sect_name(StripWhitespace(line + 1, line2));
+				if (this->sections.count(sect_name) == 0) this->sections.emplace(sect_name,
+						std::unique_ptr<ConfigSection>(new ConfigSection(*this, sect_name)));
```
Alberth289346: This looks very scary, let's keep single lines without braces a single line.

In other words stick a `{` and `}` in there when the line wraps.

src/config_reader.cpp
```diff
@@ -224,14 +185,14 @@ const ConfigSection *ConfigFile::GetSection(const char *sect_name) const
...
+std::string ConfigFile::GetValue(const std::string &sect_name, const std::string &key) const
 {
 	const ConfigSection *sect = this->GetSection(sect_name);
-	if (sect == nullptr) return nullptr;
+	if (sect == nullptr) return std::string();
 
 	const ConfigItem *item = sect->GetItem(key);
-	if (item == nullptr) return nullptr;
-	return item->value;
+	if (item == nullptr) return std::string();
```
Alberth289346, line 217: We're hiding lack of an item now, as it becomes indistinguishable from an empty value.
Do we use that feature?

src/gui_graphics.cpp
```diff
@@ -219,13 +217,14 @@ static uint8 *GetSingleLine(uint8 *text, int max_width, int *width)
  */
 void GetMultilineTextSize(StringID strid, int max_width, int *width, int *height)
 {
-	uint8 buffer[1024]; // Arbitrary max size.
-	DrawText(strid, buffer, lengthof(buffer));
+	/* \todo This design is ugly. Fix the utility function GetSingleLine to work on std::string with a srart index instead of mutable char* . */
```
Alberth289346:
```suggestion
	/* \todo This design is ugly. Fix the utility function GetSingleLine to work on std::string with a start index instead of mutable char* . */
```

src/language.cpp
```diff
@@ -128,14 +127,25 @@ StringParameters::StringParameters()
 	this->Clear();
 }
 
+/**
+ * Ensure that this Parameters object can hold at least a certain number of parameters.
+ * @param num Minimum number of parameters.
+ */
+void StringParameters::ReserveCapacity(const int num)
```
Alberth289346: Use `num_params` or so to make the meaning of the variable more intuitive.

src/language.cpp
```diff
@@ -128,14 +127,25 @@ StringParameters::StringParameters()
...
+{
+	assert(num > 0);
```
Alberth289346: using `0` as minimum looks like a valid value to me, if you don't want that, state the condition of minimum more clearly in the documentation.

src/language.h
```diff
@@ -104,16 +103,17 @@ class TextString {
...
 	 */
-	const uint8 *GetString() const
+	std::string GetString() const
 	{
-		if (_current_language < 0 || _current_language >= LANGUAGE_COUNT) return (uint8 *)"<out of bounds>";
+		if (_current_language < 0 || _current_language >= LANGUAGE_COUNT) return "<out of bounds>";
 		if (this->languages[_current_language] != nullptr) return this->languages[_current_language];
-		if (this->languages[LANG_EN_GB] != nullptr) return this->languages[LANG_EN_GB];
-		return (uint8 *)"<no-text>";
+		if (this->languages[LANG_EN_GB       ] != nullptr) return this->languages[LANG_EN_GB];
```
Alberth289346: Why the additional space before `]` ?

Alberth289346:
So much detail that gets changed!

src/config_reader.cpp
Noordfrees: It's not used, but I added functions `HasValue`/`HasItem` to allow distinguishing between non-existent and explicitly empty values.

src/language.h
Noordfrees: accidental change, reverted

src/language.cpp
Noordfrees: If 0 was allowed we'd need to ad an `assert(num > 0)` in every function that calls this one because parameter indices are 1-based. Added a note to the documentation.

PR #216, opened 2021-09-05
Add clang-tidy CI

Noordfrees:
Another handy workflow to tell us how many coding errors we have ;)
This picks up a __lot__ of problems not caught by the checkers we already have (currently 12512 hits, excluding some stylistic nitpicking we should ignore), cleaning them all up a few checks at a time is a good long-term goal re #212.

The python scripts are taken from [Widelands](https://github.com/widelands/widelands/blob/e92ffee2bd08f5a9006dccc37d02e20c042ec2e5/utils/) with only minor changes; the main script in turn is based on a template from llvm/clang.


Alberth289346:
Haven't seen @LordAro  for some time now. As this is just checking code, I am fine with merging it.
What do you think, should we wait?

PR #217, opened 2021-09-10
Finances Adjustments

Noordfrees:
Building paths and modifying the landscape now costs some money and is forbidden while the game is paused. Staff wages were reduced, though not as ridiculously cheap as in the original game.


src/path_build.h
```diff
@@ -10,6 +10,13 @@
 #ifndef PATH_BUILD_H
 #define PATH_BUILD_H
 
+static const Money PATH_CONSTRUCT_COST_FLAT         (1200);  ///< How much it costs to build a single section of flat path.
+static const Money PATH_CONSTRUCT_COST_RAMP         (1400);  ///< How much it costs to build a single section of ramped path.
+static const Money PATH_CONSTRUCT_COST_ELEVATED_FLAT(3200);  ///< How much it costs to build a single section of elevated flat path.
+static const Money PATH_CONSTRUCT_COST_ELEVATED_RAMP(3800);  ///< How much it costs to build a single section of elevated ramped path.
+static const Money PATH_CONSTRUCT_COST_CHANGE       ( 400);  ///< How much it costs to change the type of a single existing path segment.
+static const Money PATH_CONSTRUCT_COST_RETURN       (-800);  ///< How much money deletion of a single path segment returns.
```
Alberth289346:
```suggestion
static const Money PATH_CONSTRUCT_COST_RETURN       (-800);  ///< How much it costs to delete a single path segment.
```
The comment describes the unit of `PATH_CONSTRUCT_COST_RETURN` value. Live gets much simpler if you have the same unit everywhere and you seem to be doing that. The fact that its value is a negative cost so you actually return money doesn't change the unit of the value.

Compare:
`The forward speed of the car is currently -10` means it's currently driving backwards while `The backwards speed of the car is currently -10` means that the car is actually driving forward.

You start with the direction indicated in the textual description of the unit of the speed, and then change direction for the actual movement based on the sign of the value.

src/terraform.h
```diff
@@ -12,6 +12,8 @@
 
 #include <map>
 
+static const Money TERRAFORM_UNIT_COST(40);  ///< The cost for applying one elemental unit of terraforming modifications.
```
Alberth289346: At some point in the future you'll want to customize these, one idea is to collect them now all in one file or even ask for them using a function.
It's not much additional work now hopefully, and it makes customization simpler.

Alberth289346:
I don't remember whether we already do this, but OpenTTD has world change functions (a function where you change the world in some way, sum the cost of doing that, and return the cost or an error).

The advantage is that keeping computed costs correct wrt made changes is simpler. As an additional trick, it also takes a flag where you disable making the actual change. That changes the function into a test. You can run it, and get an error or check that the player has enough money for the operation. If all is fine you can run the same function again but now with enabling of making world changes.

They also use this in multi-player, where the calls to world-change functions are first transmitted to everybody (after a local check for being feasible to reduce traffic). Then at all systems you can call the function again, verify it's still allowed (it should or something is very wrong), and thus distribute world changes across all systems in the game.

It may be useful to keep something like this in the back of your mind.

Noordfrees:
Changed. Such a test mode is actually easy to implement so I did that as well now; holding down Shift displays the expected cost without applying the change.

PR #220, opened 2021-09-13
Fix queuing guests deadlocking each other

Noordfrees:
Guests don't always walk exactly at the center of a queue path, so at corners there are occasionally two guests who both think the other one is queuing in front of them, deadlocking each other. This branch fixes the bug by letting each guest remember which person they're queuing behind and allowing a guest who detects a cyclic dependency to push in front to break the deadlock.

### DIFF
--- a/src/person.cpp
+++ b/src/person.cpp
@@ -951,7 +951,7 @@ RideVisitDesire Person::ComputeExitDesire(TileEdge current_edge, XYZPoint16 cur_
 			case EDGE_SE: tile_edge_pix_pos.y = 255; break;
 			default: NOT_REACHED();
 		}
-		if (this->GetQueuingGuestNearby(original_cur_pos, tile_edge_pix_pos, false) != nullptr) {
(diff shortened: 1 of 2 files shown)


Alberth289346:
What worries me somewhat is what happens when a guest leaves the queue. Does the next guest get some notification about it?

Another possible problem is that 3 or more guests may wait on each other. Obviously that is more rare than 2, so it may not happen often. One option is to follow the chain and check the guest itself isn't in it, but that may be a long chain all the way to the first person. Perhaps first start with this and then see if we get other problems?

Alberth289346:
Code looks fine.

Noordfrees:
Cyclic dependencies with more than two guests should be an extremely rare case but I extended the fix to include this anyway.

> What worries me somewhat is what happens when a guest leaves the queue. Does the next guest get some notification about it?

Every guest checks in every animation round (i.e. every 30 ms gametime) whether there's another guest queuing in front of him. If so, the guest does nothing during this round, except if there's a cyclic dependency. Otherwise the guest starts moving forward and checks again 30 ms later. So if a guest leaves the queue, anyone waiting behind will notice by themselves within at most 30 ms.

PR #221, opened 2021-09-17
Implement Loans

Noordfrees:
Allow taking and repaying loans via the finances manager Gui. Some tweaks to the default scenario setup to ensure it's viable. Loan interest is charged daily; this also applies to staff wages now.


Alberth289346:
Code looks fine to me.

I am a bit weary about daily staff payment, it flattens out the monthly dip you normally get at the start of a new month. This "sudden" loss of money does make the game more challenging, can you buy that last piece of the ride just in time?

That said, the more steady flow is perhaps useful for users that want a more predictable flow of money. I can see a benefit there as well.

I'd say keep it in mind, let's go with the current solution for now.

PR #219, opened 2021-09-13
Refactor key navigation handling

Noordfrees:
This adds a lot of new keyboard shortcuts and replaces some of the existing ones. The full list of shortcuts is added to the manual. The backend changes are mainly to ensure that keys with modifiers are handled correctly.

PR #222, opened 2021-09-17
Fix modifier handling for Shift+Textinput

Noordfrees:
The modifier handling from #219 overgenerates a bit: Holding down Shift with no other modifiers while typing text needs to be handled the same way as when no modifier is down.


Alberth289346:
Haha, who needs uppercase letters!  :D

PR #229, opened 2022-01-15
Fix coaster station tracking across track end/start

### LINKED ISSUE #228: Crashes when roller coaster is about to arrive
From time to time, when a roller coaster train is about to arrive at the station, the game crashes with

~~~
NOT_REACHED triggered at line 1497 of ~/FreeRCT/src/coaster.cpp
Aborted
~~~

Is there some work around for this?

Noordfrees:
Fix #228
Does not fix already broken savegames though.


fnetX:
Thank you. I tested the PR and the issue appears to be gone. I removed the last track and placed it again, that seemed to have the desired effect as far as I understand the changes and your comment about already saved files.

Alberth289346:
Looks fine to me.

PR #230, opened 2022-01-16
Fix Load/Save GUI Scrollbar

Noordfrees:


Alberth289346:
Nice, I missed the scroll-offset apparently.

PR #234, opened 2022-02-20
Port FreeRCT to WebAssembly

Noordfrees:
This allows compiling FreeRCT as a WebAssembly (WASM) program that can be run in a browser. 
To enable, call CMake with `-DWEBASSEMBLY=ON`. The game can then be played by opening `<builddir>/bin/freerct.html` in the browser. 

Emscripten is required to build WASM projects. You'll need the latest version, the one from the current Ubuntu repos will *not* work.

This feature should be considered experimental as it is barely tested; in particular savegames do not survive a page refresh as WASM programs have no access to the computer's filesystem for safety reasons.

When this is merged I'll write an integration wrapper for the website, making it possible to try out the game online without having to install it locally.


src/video.h
```diff
@@ -61,7 +61,9 @@ class VideoSystem {
 	std::string Initialize(const std::string &font_name, int font_size);
 	bool SetResolution(const Point32 &res);
 	void GetResolutions();
-	void MainLoop();
+	static void MainLoop();
+	static bool MainLoopCycle();
```
Alberth289346: Wouldn't it be useful to keep the static methods as a separate list from the non-static methods?

EDIT: Maybe also state such methods as static in the .cpp file in some way to avoid interpreting the code in the wrong way. One way would be to add `/* static */` before the return type, but other ways are also possible.

Otherwise looks fine, please consider and feel free to merge when done.

Noordfrees:
Thanks for the review. The WASM port is now up for in-browser playing at https://freerct.net/play/play?guest=true#anchor
I've configured it to pull and compile the latest FreeRCT version nightly. 
Most of the time the game runs smoothly, but sometimes it stutters a lot, I'm still investigating why that happens.

PR #237, opened 2022-03-01
Allow moving scrollbars with the mouse wheel

Noordfrees:


Alberth289346:
Not entirely sure how much control you have when there are several scrollbars available, but we may not have that problem yet.

PR #238, opened 2022-03-01
Fix CodeQL Hits

Noordfrees:


src/rcdgen/nodes.cpp
```diff
@@ -1027,8 +1027,8 @@ int TIMABlock::Write(FileWriter *fw)
 	FileBlock *fb = new FileBlock;
 	fb->StartSave(this->blk_name, this->version, 16 + (8 * this->frames) - 12);
 	fb->SaveUInt32(this->frames);
-	for (int8 f = 0; f < this->frames; f++) fb->SaveUInt32(this->durations[f]);
-	for (int8 f = 0; f < this->frames; f++) {
+	for (int f = 0; f < this->frames; ++f) fb->SaveUInt32(this->durations[f]);
```
Alberth289346, line 1030: Pre-increment vs post-increment makes no difference whatsoever.

It comes from the idea that compilers literally do what you say, which hasn't been true for the last 30 years or so at least.

Even then, in `f++;` the `f` value is not externally used, so the entire supposedly disadvantage of needing to subtract 1 isn't even there.

PR #243, opened 2022-03-13
Fix packaging warnings and errors flagged by debian's lintian

### LINKED ISSUE #242: Lintian warnings/errors for Debian builds
I've been looking into the requirements for getting the DEB packages built by our CI into the debian repos, and it turns out we have a lot of major quality issues that will need to be fixed – lack of a debian-style changelog and copyright file, lack of Release-configured builds, install to /bin and /usr/share instead of /usr/games and /usr/share/games, and a number of other nits.

I'd like to address this for v0.1, so with at least one other release-critical new issue (#240) discovered as well I've postponed the expected release date by a week to give us some more time.

Noordfrees:
Main changes:
- You can now build a release build with the CMake option `-DRELEASE=ON`.
- You can decide whether the file structure generated by `make install` is that of a user-local or flatpak installation (~/.local/bin, ~/.local/share) or that of a system-wide Debian/Ubuntu installation (/usr/games, /usr/share/games) with `-DDEBIAN_PACKAGING=ON` (system-wide) or OFF (local, that's the default). The Ubuntu CI builds debian-style packages; the Windows and Mac CIs build non-debian packages.
- As a side-effect, this fixes a bug in our CMakeLists that prevented some compiler settings from taking effect. The compiler now emits many new warnings which should be cleaned up sometime.

---

There are a handful of non-critical lintian warnings left:
- `W: freerct: extended-description-line-too-long line 1`
- `W: freerct: no-manual-page usr/games/freerct`
- ~`W: non-standard-dir-perm 0775 != 0755` for every directory~ edit: doesn't seem to happen with the CI builds

---

With those changes, we can request inclusion of the package in the Debian repos after the release.

Fixes #242


Newbytee:
As-is, this causes the Flatpak to crash on launch with "Insufficient graphics loaded.". I'll look into it.

Newbytee:
The Flatpak doesn't really like that the stuff gets installed to `/usr/share/games`. Prior to this PR, it would get installed to `/app/share` and `/app/bin`, which is where Flatpak wants it. Would it be possible to hide this new distribution install behaviour behind a switch? Maybe something like `DIST_PACKAGING=Debian`. E.g. Fedora also doesn't put games in `/usr/share/games`.

Noordfrees:
I renamed the flag to `DEBIAN_PACKAGING`; when it's left `OFF` the flatpak should behave the same way as in master…

CMakeLists.txt
```diff
@@ -22,6 +42,23 @@ IF(NOT DEFINED FREERCT_BASENAME)
 	set(FREERCT_BASENAME "net.freerct.FreeRCT")
 ENDIF()
 
+IF(DEFINED FREERCT_RELEASE)
+	set(RELEASE ON)
```
Newbytee:
```suggestion
IF(DEFINED RELEASE)
```

CMakeLists.txt
```diff
@@ -9,11 +9,31 @@
 project(FreeRCT)
 cmake_minimum_required(VERSION 2.8)
 
+# Uncomment and adjust this for release branches:
+# set(FREERCT_RELEASE "0.1")
```
Newbytee:
```suggestion
set(FREERCT_RELEASE "0.1")
```

Newbytee (changes requested):
Personally I would prefer it better like this since it means releases don't need patching to become releases.

Noordfrees:
We'll need to create a separate branch for every release in any case so we can give them a proper version string `0.1` instead of `0.1~git2345-0.1`. But I realized now that `FREERCT_RELEASE` as I added it here doesn't really make sense…

CMakeLists.txt
```diff
@@ -9,11 +9,28 @@
 project(FreeRCT)
 cmake_minimum_required(VERSION 2.8)
 
+# Add a compiler flag VAR_NAME=FLAG. If VAR_NAME is already set, FLAG is appended.
+FUNCTION(add_flags VAR_NAME FLAG)
+	IF(${VAR_NAME})
```
Newbytee:
```suggestion
	IF(DEFINED VAR_NAME)
```
Isn't this more proper? I'm unsure if I understand this right.

Noordfrees: Good catch, fixed. However VAR_NAME is a variable containing the name of the variable to modify, so we still need to dereference it in order to check if the variable exists.

Newbytee (approved):
Looks good to me now!

PR #245, opened 2022-03-13
Create parent directories in `MakeDirectory`

### LINKED ISSUE #240: Description of MakeDirectory function is incorrect
MakeDirectory's description claims that it creates all parents if they don't already exist (see https://github.com/FreeRCT/FreeRCT/blob/master/src/fileio.cpp#L262). However, this is false. If the parents don't exist, `mkdir` will fail. I'm not sure if the same applies to `CreateDirectory` (the equivalent Win32 API), but my brief research indicates that it may be the same there. This, in turn, causes the Flatpak to crash on startup with `Failed creating directory '/home/<username>/.var/app/net.freerct.FreeRCT.Devel/data/freerct/save' (No such file or directory)`, so this is currently a blocker for submitting the Flatpak. It may also affect other types of packages, but I didn't test anything like that.

Noordfrees:
Fix #240

### DIFF
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -260,19 +260,23 @@ bool RcdFileReader::GetBlob(void *address, size_t length)
 }
 
 /**
- * Create a directory and all its parents if it did not exist yet.
+ * Create a directory and all its parent directories if it did not exist yet.
  * @param path Path of the directory.
- * @todo At the time of writing (2021-06-30) this is needed and tested only on Linux. Before using it anywhere else, test this on all platforms (especially Windows).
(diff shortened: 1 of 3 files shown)


Newbytee:
Is there any case where we don't want to create parents when making a directory?

Newbytee:
Other than that, looks good to me. I can confirm the Flatpak works without issues with this patch applied.

Newbytee:
I think this looks good now. Do you expect me to merge it? :thinking:

Newbytee (approved):

PR #247, opened 2022-03-13
Staff wage cuts

Noordfrees:
With the current balancing settings it is practically impossible to run your park efficiently for any extended time. The main reason are the staff wages. In RCT1 a staff member earns around 50$ a month which is a ridiculous amount, but since you can't truly play with more realistic wages this PR reduces salaries by 90% allowing you to actually turn a profit now.


Newbytee (approved):
Looks good to me, but maybe you want more review before merging this?

PR #250, opened 2022-03-14
Add the 0.1 release to the metainfo

Newbytee:
Needs to be here before we make the tag/branch.


Newbytee:
@Noordfrees Can I merge this? With this merged the Flathub build should succeed if your webserver changes worked too.

PR #252, opened 2022-03-18
WebAssembly integration enhancements

Noordfrees:
To make the online version run more smoothly, and fix that it occasionally breaks due to a target dependencies ambiguity.

Although this *should* affect only WASM builds, merging should be held off until after the release to prevent any chance of last-minute damage.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -11,8 +11,11 @@ ENDIF()
 
 project(freerct)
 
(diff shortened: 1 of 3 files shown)


Alberth289346:
I typically make such pull requests "draft" so they are not merged accidentally.

Newbytee:
Is this ready to be merged now that v0.1 has been tagged and released?

PR #253, opened 2022-03-18
Fix g++ compiler warnings

Noordfrees:
Mostly commenting out unused parameters and adding casts to signed-unsigned comparisons, plus some (required) refactoring to make `Point` and `Rectangle` constructors `constexpr` with initializer lists (something I intend to have more of re #212).

*There are more warnings in the CI logs, this is just the first batch…*


Alberth289346:
Patch looks fine to me, not sure if you want to add more. Feel free to merge if you consider it done.

Some suggestions to consider:

## Unused variables

I looked for an alternative way to express 'unused', and gcc has a variable attribute for this: https://gcc.gnu.org/onlinedocs/gcc-9.3.0/gcc/Variable-Attributes.html#Variable-Attributes  The unused notion is in the common variable attributes as `unused`.

That would lead to
``` cpp
int main(__attribute__((unused)) int argc, char __attribute__((unused)) *argv[]) { return 0; }
```
or with a macro
``` cpp
#define UNUSED __attribute__(unused))
int main(UNUSED int argc, char UNUSED *argv[]) { return 0; }
```
This would also be an option but it looks a bit too much magical I think.
``` cpp
#define UNUSED(DECL) __attribute__((unused)) DECL

int gety(UNUSED(int x), int y) {
    return y;
}
```

A simpler option is perhaps
``` cpp
#define UNUSED(V)

int gety(int UNUSED(x), int y) {
    return y;
}
```
Complicated types will look funny though `int *UNUSED(x)[]`

EDIT: This may throw off some other tools, that don't perform macro processing and look at function declarations (like the documentation tool for example). Your choice is more safe in that respect. Also, you'll need similar solutions for other compilers eventually if you use the gcc specific solution.

Last but not least, C++17 seems to get support for this notion, so perhaps just wait a while longer?

## Signed versus unsigned

You're now making local fixes to make the compiler happy at each spot. That will give an X% of the code using unsigned data, and an (100-X)% of the code using signed data.

I am wondering if it would be simpler in the long run to make some global decisions about types in the code. That would probably simplify the types being used. You could decide for example that bit-encoded data is always unsigned (it probably already is), counters and sizes are always signed (currently they are not since sizeof() is unsigned)), positions are always signed (probably the best option), and so on.

Noordfrees:
Re unused variables: I don't like compiler-specific solutions because there are so many situations which would either break or require different handling, and we support more compilers than just g++ so we'd need at least an equivalent solution for clang and probably more as well. 
I do like the look of this C++17 `[[maybe_unused]]` attribute which would be much cleaner. Do you think it would make sense for us to migrate to C++17 or even C++20 now, or how long should we wait before we do?

Re signed/unsigned: I don't think we can fully avoid this because C++'s builtin `size_t` is unsigned, so any comparison with `some_container.size()` in it is always likely to cause this warning, especially if we make more variables `signed`. But I think it's not much of a problem to add a `static_cast` here and there; the majority of cases is `signed` anyway because of implicit conversions to `signed int`.

Alberth289346:
> Re unused variables: I don't like compiler-specific solutions ...

Yep, I can imagine that. I am not sure I like the macro thing enough either to switch to it.

With respect to C++17, you at least need a compatible compiler for all supported platforms.

If you want many users being able to compile the code, you should wait until having C++17 is common for all platforms. For Linux that means the major LTS distributions should support that C++ version, don't know how to measure it for other platforms.

> Re signed/unsigned: I don't think we can fully avoid this because C++'s builtin size_t

Indeed, but a standard pattern could be always add a static cast around it, or even have a macro for it (a normal function is likely not working properly in all cases).
As with all code style, the main point of such rules is to make code more predictable so you make less errors.

Noordfrees:
In that case I think we'd be pretty safe switching to C++17 though not 20. Ubuntu Bionic has g++7 which [fully supports](https://en.cppreference.com/w/cpp/compiler_support/17) 17 but not 20. [Most other distros](https://repology.org/project/gcc/versions) also seem to have gcc >=7 in their ± recent versions. Windows build environments usually have the cutting-edge version of pretty much everything (they're nearly always the first CI that breaks when a library is updated :wink:) so if our CI can handle it it should be good. No idea about MacOS but besides a CI that checks that it compiles we don't currently support macs.

I'll add a commit to this branch to update the C++ standard version and use `[[maybe_unused]]`, let's see what the CI makes of this. - Edit: No issues.

---
> > Re signed/unsigned: I don't think we can fully avoid this because C++'s builtin size_t
> Indeed, but a standard pattern could be always add a static cast around it, or even have a macro for it (a normal function is likely not working properly in all cases).

Good point, +1 then for a wrapper and using signed types for everything except bitsets. The more modern (and safer) way are template functions (which reminds me that C++17 also has better templates support, yippieh!), so something like
```c++
template<typename T> int64 Size(const T &t) { return t.size(); }
```
used as `Size(items)` instead of `items.size()` should do the trick.
But that will be another large refactoring for a separate branch.

src/fixed_ride_type.cpp
```diff
@@ -99,7 +99,7 @@ int FixedRideInstance::EntranceExitRotation(const XYZPoint16& vox) const
 	NOT_REACHED();  // Invalid entrance/exit location.
 }
 
-void FixedRideInstance::GetSprites(const XYZPoint16 &vox, uint16 voxel_number, uint8 orient, const ImageData *sprites[4], uint8 *platform) const
+void FixedRideInstance::GetSprites(const XYZPoint16 &vox, uint16  /* voxel_number */, uint8 orient, const ImageData *sprites[4], uint8 *platform) const
```
Alberth289346: Missed one, or is this intentional?

EDIT:
```suggestion
void FixedRideInstance::GetSprites(const XYZPoint16 &vox, [[maybe_unused]] uint16 voxel_number, uint8 orient, const ImageData *sprites[4], uint8 *platform) const
```

src/geometry.h
```diff
@@ -234,33 +236,34 @@ struct Rectangle {
...
 	 * Copy constructor.
 	 * @param rect %Rectangle to copy.
 	 */
 	template<typename PT2, typename SZ2>
-	Rectangle(const Rectangle<PT2, SZ2> &rect)
+	constexpr Rectangle(const Rectangle<PT2, SZ2> &rect)
+	:
+		base  (rect.base  ),
+		width (rect.width ),
+		height(rect.height)
```
Alberth289346: You are aware that maintaining vertical alignment does take a lot of effort?
Change one name, and the entire column needs fixing.

I do like vertical alignment, but these look a bit too simple to have much benefit to me.
I'd probably even stick lines 253 to 256 all behind 252, but that may be me :)

src/minimap.cpp
```diff
@@ -103,7 +103,7 @@ void Minimap::UpdateButtons()
  */
 Point32 Minimap::GetRenderingBase(const Rectangle32 &widget_pos) const
 {
-	const int required_size = this->zoom * (_world.GetXSize() + _world.GetYSize());
+	const unsigned required_size = this->zoom * (_world.GetXSize() + _world.GetYSize());
```
Alberth289346, line 106: This is opposite to using signed for sizes which you seemed to discuss in the last comment

src/path_gui.cpp
```diff
@@ -834,7 +834,7 @@ bool PathBuildGui::MoveSelection(bool move_forward)
 	return false;
 }
 
-void PathBuildGui::OnChange(ChangeCode code, uint32 parameter)
+void PathBuildGui::OnChange(ChangeCode code, uint32  /* parameter */)
```
Alberth289346:
```suggestion
void PathBuildGui::OnChange(ChangeCode code, [[maybe_unused]] uint32 parameter)
```

src/people.cpp
```diff
@@ -372,7 +372,7 @@ void Guests::AddFree(Guest *g)
  */
 Guest *Guests::GetFree()
 {
-	bool b = this->FindNextFreeGuest();
+	[[maybe_unused]] bool b = this->FindNextFreeGuest();
```
Alberth289346: lol :) 

Maybe also use this for the `#ifndef NDEBUG` things? Let the compiler figure out if it's useful.

This solution may have the dis-advantage of getting old assignments in the code that are never used due to code using it has been removed in the past.

src/ride_gui.cpp
```diff
@@ -282,7 +282,7 @@ void RideSelectGui::SetNewRide(int number)
 	this->current_ride = -1;
 	number = std::min(number, this->ride_types[this->current_kind] - 1);
 	if (this->ride_types[this->current_kind] > 0) {
-		for (int i = 0; i < _rides_manager.ride_types.size(); i++) {
+		for (unsigned i = 0; i < _rides_manager.ride_types.size(); i++) {
```
Alberth289346, line 285: Switches to unsigned

Alberth289346:
Looks good to me, except for a few missed non-changed parameters. Feel free to merge after fixing.

PR #256, opened 2022-03-20
Monthly Autosave

### LINKED ISSUE #248: Autosaves
I think autosaving the games is desirable and better than asking the user to save periodically.

I think a good solution would be to ask the user for a name upon starting a new game and automatically saving the game in certain intervals, e.g. every 1,3,5 minute(s).

Consider always keeping one backup copy from the previous save in case some bug happens, it might help triaging / repairing the file.

Noordfrees:
Fixes #248

### DIFF
--- a/src/gamecontrol.cpp
+++ b/src/gamecontrol.cpp
@@ -22,6 +22,7 @@
 #include "weather.h"
 #include "freerct.h"
 #include "fileio.h"
+#include "rev.h"
 
 GameModeManager _game_mode_mgr; ///< Game mode manager object.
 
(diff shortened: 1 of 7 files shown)


src/gamecontrol.cpp
```diff
@@ -34,6 +35,7 @@ void OnNewYear()
 /** Runs various procedures that have to be done monthly. */
 void OnNewMonth()
 {
+	Autosave();
```
Alberth289346, line 38: If I load the save again, will it perform the other `OnNewMonth` updates?

Noordfrees: The save isn't performed immediately – `Autosave()` delegates to `SaveGame()` which only sets the game controller's `next_action` variable to `GCA_SAVE_GAME`, which is then carried out on the next frame. So all the monthly updates will be performed before the save, and the saved game will start one tick into the new month.

PR #257, opened 2022-03-20
Use initializer lists in constructors

Noordfrees:
Re #212

### DIFF
--- a/src/coaster.cpp
+++ b/src/coaster.cpp
@@ -94,7 +94,6 @@ bool CarType::Load(RcdFileReader *rcdfile, const ImageMap &sprites)
 
 CoasterType::CoasterType() : RideType(RTK_COASTER)
 {
-	this->voxels = {};
 }
 
 CoasterType::~CoasterType()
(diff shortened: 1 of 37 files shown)


Alberth289346:
While initializers is the recommented practice, do you know if it makes any difference code-wise?
Seems to me it would generate the same code, wouldn't it?

Changes themselves are good (as in, don't seem to make any change in initial value) but I haven't detected the rule for formatting the initializers. It looks like there are 3 or so different forms?

Noordfrees:
For primitive types there shouldn't be a difference, but class/struct instances are implicitly default-constructed if they don't have an entry in the initializer list. So initializing a `Point` or `Rectangle` in the old-style way first calls the default constructor and then the assignment operator with a reference to another temporary object, resulting in at least an extra copy plus potentially other side effects from custom operators. See https://stackoverflow.com/q/4289899

For style, the rule I applied is:
1. Try to fit it all on a single line:
```c++
Foo::Foo() : bar(1)
{
}
```
2. If that's too long or unreadable, try to fit the initializers on one separate line:
```c++
Foo::Foo(int i, int j, int k) : public Bar(i, j, k),
hello(1), world(2), another(3), variable(4)
{
}
```
3. If that's also too long, one initializer per line:
```c++
Foo::Foo()
:
	hello(1),
	world(2),
	another(3),
	variable(4),
	with_a_very_very_long_name_that_would_otherwise_overflow_the_line(5)
{
}
```
The parent constructor call should always stay on the same line as the function header to distinguish it from the actual initializers.

It is also worth noting that default constructor calls (both for member variables and for the parent class constructor) can be omitted. In some cases I chose to include them to highlight that these member variables are used by other initializers (in which case the compiler will print a warning if the order of initialization is incorrect), but where this is not the case it is safe to leave them out.

Alberth289346:
SO quotes the compiler manual (which is of course correct as that is the definition), but in my experience, actually generated code may skip, shuffle or remove parts that have no use. The language standard doesn't define what a compiler should literally do, it defines behavior by "must behave in the same way as when you would do .....". In other words, a compiler can do anything it likes as long as it cannot be distinguished externally from the sequence defined in the standard.

No compiler will initialize the same data multiple times even if the language standard says it should (unless you have turned off all optimization or have code that is not side-effect free or may throw exceptions). For the usual simple cases that we have I'd expect not much difference.

PR #261, opened 2022-03-23
Add Haiku support

win8linux:
This adds support for building FreeRCT on Haiku, through adding platform-specific directories for SDL2.

### DIFF
--- a/CMake/FindSDL2.cmake
+++ b/CMake/FindSDL2.cmake
@@ -74,12 +74,13 @@ SET(SDL2_SEARCH_PATHS
   /opt/local # DarwinPorts
   /opt/csw # Blastwave
   /opt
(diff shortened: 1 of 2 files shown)


Noordfrees (approved):
LGTM, and thanks for porting FreeRCT to Haiku :)

augiedoggie:
I think this needs some cleanup, but I haven't tested it.  It's not usually necessary to add the lib directory and the include paths just need to have 'SDL2' added, as is done with other Haiku ports like [cdogs-sdl](https://github.com/haikuports/haikuports/blob/c4dbd3e3c72e20963a991281fdaec8ceba8bdf12/games-arcade/cdogs_sdl/patches/cdogs_sdl-1.2.0.patchset#L16)

Issue #260, opened 2022-03-23
Building for MacOS


Flipp3rrr:
I noticed MacOS is not officially supported, so I tried compiling it and it turns out all dependencies can be installed with Homebrew (`$ brew install flex bison libpng sdl2 sdl2_ttf cmake make`) after which a functional executable can be produced (`cmake && make && make install`).

I've made a fork ([Flipp3rrr/FreeRCT](https://github.com/Flipp3rrr/FreeRCT)) at first, but since it's quite trivial i'm unsure if I should submit a pull request.

System specifications:
- MacOS Big Sur 11.6.1
- 1,1 GHz Quad-Core Intel Core i5
- 8 GB 3733 MHz LPDDR4X
- Intel Iris Plus Graphics 1536 MB

I've also added a screenshot of FreeRCT running under MacOS Big Sur below:

<img width="912" alt="image" src="https://user-images.githubusercontent.com/33151653/159685516-c7df1991-4883-468f-acc6-89fb9547ae6b.png">

Flipp3rrr:
Saving & loading appears to be broken, but this might be a bug in FreeRCT itself.

Loading a previously saved game caused the following crash:

```sh
$ freerct
ERROR: Loading '/Users/flipp3rrr/.local/share/freerct/save/Park Test.fct' failed: Missing pattern name for cstc
Assertion failed: (anim_type >= ANIM_BEGIN && anim_type <= ANIM_LAST), function GetAnimationSprite, file /Users/flipp3rrr/Documents/GitHub/FreeRCT/src/sprite_store.h, line 656.
[1]    48035 abort      freerct
```

This is what the park in my original message looked like after loading it as a saved game:

<img width="912" alt="image" src="https://user-images.githubusercontent.com/33151653/159688724-db886c46-7c3a-4d57-ab2c-5f3c824cb60c.png">

Closing the game with the red `x` on the top left also caused the following crash:

```sh
$ freerct
ERROR: Loading '/Users/flipp3rrr/.local/share/freerct/save/Park Test.fct' failed: Missing pattern name for cstc
Assertion failed: (voxel), function RemoveFromWorld, file /Users/flipp3rrr/Documents/GitHub/FreeRCT/src/track_piece.cpp, line 178.
[1]    49386 abort      freerct
```

Here's the savegame file:

[Park Test.fct.zip](https://github.com/FreeRCT/FreeRCT/files/8332562/Park.Test.fct.zip)

[...]

PR #262, opened 2022-03-24
Add build instructions for MacOS

Flipp3rrr:
Let me know if any testing has to be done on MacOS.


Flipp3rrr:
See also: issue #260

PR #268, opened 2022-03-28
Startup fixes for clean installations

### LINKED ISSUE #267: Can't run FreeRCT like described in the readme
I followed the instructions to download and compile FreeRCT, but trying to [run the game like described](https://github.com/FreeRCT/FreeRCT#running-the-program) doesn't work:

~~~~
$:> cd bin/
$:> ./freerct 
Failed creating directory '/home/kaputtnik/.local/share/freerct/save/' (File exists)
~~~~

The directory was created by the program, so i wonder why the message appears and exits. Next try:

~~~~
$:> ./freerct 
Data file data/font/Ubuntu-L.ttf is missing, the installation seems to be broken!
~~~~

Hm.... probably because i run from the `build/bin/` directory? The `data` folder is in the base directory (`FreeRCT/data/` after git clone) so i tried to run the binary from there:

~~~~
cd ../../
$:> build/bin/freerct 
Insufficient graphics loaded.
~~~~

Hm... maybe i have to use the `-u` switch:

~~~~
$:> cd build/bin/
$:> ./freerct -u /home/kaputtnik/Quellcode/FreeRCT/data/
Failed creating directory '/home/kaputtnik/Quellcode/FreeRCT/data//save/' (File exists)
~~~~

Ok, i know this message, next try:
~~~~
$:> ./freerct -u /home/kaputtnik/Quellcode/FreeRCT/
Data file data/font/Ubuntu-L.ttf is missing, the installation seems to be broken!
~~~~

And this also don't work for me:
~~~~
cd ../
$:> make run
Consolidate compiler generated dependencies of target rcdgen
[ 15%] Built target rcdgen
[ 23%] Built target rcd
Consolidate compiler generated dependencies of target freerct
[100%] Built target freerct
Data file data/font/Ubuntu-L.ttf is missing, the installation seems to be broken!
make[3]: *** [CMakeFiles/run.dir/build.make:70: CMakeFiles/run] Fehler 1
make[2]: *** [CMakeFiles/Makefile2:160: CMakeFiles/run.dir/all] Fehler 2
make[1]: *** [CMakeFiles/Makefile2:167: CMakeFiles/run.dir/rule] Fehler 2
make: *** [Makefile:189: run] Fehler 2
~~~~

What do i miss?

OS: archlinux

Noordfrees:
Fixes #267

One commit to go a few levels up and/or down when scanning for RCD and data files (limited to a few levels to avoid unnecessary in-depth searching), and one commit for the directory creation issue. 

Tested with a nonexistent custom user data dir with and without trailing slash, and running from the top-level dir and from `build` and `build/bin`.

### DIFF
--- a/src/fileio.cpp
+++ b/src/fileio.cpp
@@ -266,9 +266,8 @@ void MakeDirectory(std::string path)
 	if (path.empty() || PathIsDirectory(path.c_str())) return;
 
 	/* Strip trailing path separators. */
-	const size_t dir_sep_len = strlen(DIR_SEP);
-	while (path.size() >= dir_sep_len && path.compare(path.size() - dir_sep_len, dir_sep_len, DIR_SEP) == 0) {
-		for (size_t i = 0; i < dir_sep_len; ++i) {
+	while (StrEndsWith(path.c_str(), DIR_SEP, false)) {
+		for (size_t i = strlen(DIR_SEP); i > 0; --i) {
 			path.pop_back();
 		}
 	}
@@ -346,7 +345,7 @@ const std::string &GetUserHomeDirectory()
  */
 std::string FindDataFile(const std::string &name)
 {
-	for (std::string path : {std::string(".."), freerct_install_prefix()}) {
+	for (std::string path : {std::string("."), std::string(".."), std::string("..") + DIR_SEP + "..", freerct_install_prefix()}) {
 		path += DIR_SEP;
 		path += name;
 		if (PathIsFile(path.c_str())) return path;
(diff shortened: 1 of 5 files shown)


frankystone:
Approved testing.

Not tested if this affects an installation of freerct.

src/fileio.cpp
```diff
@@ -261,10 +261,18 @@ bool RcdFileReader::GetBlob(void *address, size_t length)
...
 {
-	if (PathIsDirectory(path.c_str())) return;
+	if (path.empty() || PathIsDirectory(path.c_str())) return;
 
+	/* Strip trailing path separators. */
+	while (StrEndsWith(path.c_str(), DIR_SEP, false)) {
+		for (size_t i = strlen(DIR_SEP); i > 0; --i) {
+			path.pop_back();
+		}
+	}
```
Alberth289346, line 273: Perhaps keep at least one character? `/////` as path isn't very sane, but at least `/` is then a technically valid end result rather than the empty path (which I am not sure what that would mean here).

Noordfrees: This should already be covered by the `PathIsDirectory()` call in line 266 – I think we can always guarantee that the root directory exists.

Alberth289346: ok

PR #270, opened 2022-07-17
Add Preload information to savegame files and the loadsave GUI

Noordfrees:
This adds the timestamp and scenario name of a savegame to the Load dialog, and also an icon to indicate whether the file was saved with the same version of FreeRCT so you can see whether there could be compatibility issues.

image


developer_documentation/data_format.rst
```diff
@@ -1743,20 +1743,23 @@ Offset  Length  Version  Field name          Description
...
- 248       4     10-     toolbar_scenery     Toolbar: Scenery GUI button.
- 252       4     10-     toolbar_terrain     Toolbar: Terraform GUI button.
- 256       4     10-     toolbar_staff       Toolbar: Staff GUI button.
- 260       4     10-     toolbar_inbox       Toolbar: Inbox GUI button.
- 264       4     10-     toolbar_finances    Toolbar: Finances GUI button.
- 268       4     10-     toolbar_objects     Toolbar: Path objects GUI button.
- 272       4     10-     toolbar_view        Toolbar: View menu GUI button.
- 276       4      2-     texts               Text of the guis (reference to a TEXT block).
- 280                                         Total length.
+ 224       4     12-     loadsave_err        Loadsave icon for files with errors.
```
Alberth289346: Above you don't delete the line with offset 224, so now two variables live at the same offset.

Alberth289346: Not sure why you don't simply add it to the bottom, but fair enough.

src/loadsave.cpp
```diff
@@ -322,7 +325,29 @@ void Saver::PutText(const std::string &str, int length)
 
 /* When making any changes to saveloading code, don't forget to update the file 'doc/savegame.rst'! */
 
-static const uint32 CURRENT_VERSION_FCTS = 10;  ///< Currently supported version of the FCTS pattern.
+static const uint32 CURRENT_VERSION_FCTS = 11;  ///< Currently supported version of the FCTS pattern.
+
+PreloadData Preload(Loader &ldr)
```
Alberth289346, line 335: No documentation?

src/loadsave.cpp
```diff
@@ -322,7 +325,29 @@ void Saver::PutText(const std::string &str, int length)
...
+	PreloadData result;
+
+	if (version >= 11) {
+		result.timestamp = ldr.GetLongLong();
+		result.revision = ldr.GetText();
+		result.scenario_name = ldr.GetText();
+	} else {
+		result.timestamp = 0;
+		result.revision = "?";
+		result.scenario_name = "?";
```
Alberth289346: This `0` is likely where `1970` is coming from in the screenshot. While technically correct, wouldn't it be nicer if you showed "not available" or (`n/a` if there is no room)  in the display instead?

Here you may want to use `0` and an empty string for that (assuming scenario names are never empty).

src/loadsave.cpp
```diff
@@ -408,6 +433,32 @@ bool LoadGameFile(const char *fname)
...
+	PreloadData result;
+	if (fname == nullptr) return result;
+
+	try {
+		FILE *fp = fopen(fname, "rb");
+		if (fp == nullptr) return result;
+		Loader ldr(fp);
+		result = Preload(ldr);
+		fclose(fp);
+	} catch (...) {
```
Alberth289346: I was thought to avoid catching all exceptions, and instead only catch those that you really want to catch.
The reason for that is here you may grab totally unrelated exceptions that arise due to programming errors, and silently discard them. That means bugs may be hiding in here without ever being found.

The only exception to that rule is if you rethrow.

Alberth289346:
Nice extension.

Issue #258, opened 2022-03-20
Translation System Refactoring


Noordfrees:
Follow-up for one of the topics that emerged in #149.

Our current translations system is not translators-friendly. I propose these steps to change that:

1. Switch to a standardised file format which we can still easily parse (preferably by writing a simple custom parser without using 3rd-party libraries, e.g. YAML or JSON, but *not* e.g. gettext's PO files).
 I currently favour YAML since it has an easily parsable syntax and is monolingual like our current translation files. But we might still decide that another format is more convenient while implementing the change.
2. Support pluralized strings. Currently we use awkward catch-all constructs like "%1% guest(s)". Some languages have very complex plural forms though (and sometimes more than 2 plurals) which we really should start to support.
3. Use a translations platform such as Weblate for translating. 
 I'm a fan of doing stuff ourselves instead of using 3rd-party integrations where reasonably possible, and Weblate offers free self-hosting with a great doc.  I'd like to look into whether we can add this to our webserver and allow all users who are registered on our website to translate the game directly from our site, with no 3rd party website involved. We'll also need a bot that frequently syncs the server's translations database with the GitHub repo (that's the easiest part).
 Should this be too difficult we can also use Weblate's hosting, which is free for open source projects.

---

Useful links:
[Weblate self-hosted install guide](https://docs.weblate.org/en/latest/admin/install.html)
[Weblate supported file formats](https://docs.weblate.org/en/latest/formats.html) (We'll need a monolingual format that supports plurals, e.g. "Ruby YAML" but *not* plain YAML.)

Alberth289346:
As an alternative, the OpenTTD language system is an option I think.
Language files have a number of header lines, and then are simple `NAME : text` lines, like
```
STR_QUANTITY_STEEL                                              :{WEIGHT_LONG} of steel
STR_QUANTITY_VALUABLES                                          :{COMMA}{NBSP}bag{P "" s} of valuables
STR_QUANTITY_COPPER_ORE                                         :{WEIGHT_LONG} of copper ore
```
It's a very rich language system, it can express gender, plural, and cases.

I wrote eints at some point, a web translator system for easier translating projects. It started with NewGRFs, and expanded itself to almost everything, including all text of the openttd program. Not sure how much got changed since I left.

Some pseudo-random links:

- "front" door https://www.openttd.org/development
- Entry point web translator https://translator.openttd.org/
[...]

PR #273, opened 2022-08-21
Translations System Update: Pluralized Strings and R-YAML Language Files

Noordfrees:
Re #258 

I decided to use R-YAML instead of the OpenTTD language system for two reasons. One, although OpenTTD's system allows a lot of flexibility, we don't *need* this flexibility. We would use only a small subset of its features, which we are already using, just in a different implementation way. There is no reason to assume we would make more use of it. So this system would add a lot of complexity (which would need to be implemented and maintained) for no expected benefit. Second, portability. YAML is a standardised format supported by nearly all text editors and a good number of translation platforms. Most text editors would not know how to provide good syntax highlighting for OpenTTD's format, and there is exactly one translation platform dedicated to this format, which is the ultimate vendor lock-in – even though it's open source, it is surely tailored to OpenTTD's needs and any customisation would require active development and maintenance from us. A solution like e.g. Weblate, which specifically aims to be integrated well into many third-party projects, is more interchangeable and will make the translations workflow more flexible and customisable in the future.

---

The other main change beside the file format here are **pluralized strings**. 
image image
The back-end to support languages with fancier plural rules like Czech or Gaelic is a bit involved, and we do not have such a language yet (at least I think we don't). I would appreciate if @Newbytee, @fnetX, and @Flipp3rrr could have a look at the pluralized strings in Swedish, French, and Dutch respectively and especially whether the plural rules are correct (I assumed they're the same as English, but I don't speak either of these languages). The Danish and Spanish translations seem unmaintained but they did not have any existing translations for now-pluralized strings.


Newbytee:
Where do I view the pluralised strings?

Alberth289346:
Plural definitions for about 53 languages:
https://github.com/OpenTTD/OpenTTD/blob/master/src/strings.cpp#L523

Noordfrees:
Thanks, I fixed the French rule, the other languages were correct.

To view pluralized strings, go the the yml file ([sv_SV](https://github.com/Noordfrees/FreeRCT/blob/i18n-ryaml/graphics/rcd/lang/sv_SE.yml) · [nl_NL](https://github.com/Noordfrees/FreeRCT/blob/i18n-ryaml/graphics/rcd/lang/nl_NL.yml) · [fr_FR](https://github.com/Noordfrees/FreeRCT/blob/i18n-ryaml/graphics/rcd/lang/fr_FR.yml)) and search the file for the key `other` which will match all pluralized strings.

e.g. https://github.com/Noordfrees/FreeRCT/blob/f7f9b564d892130572a755675c4cc70f410d76cd/graphics/rcd/lang/sv_SE.yml#L286-L288
```yaml
		MESSAGE_BAD_RATING:
			one:   "Parkbetyget är kritiskt lågt. Parken kommer stängas om %1% vecka om betyget inte förbättras."
			other: "Parkbetyget är kritiskt lågt. Parken kommer stängas om %1% veckor om betyget inte förbättras."
```

src/coaster_gui.cpp
```diff
@@ -428,9 +428,11 @@ void CoasterInstanceWindow::SetWidgetStringParameters(WidgetNumber wid_num) cons
 
 		case CIW_NUMBER_TRAINS:
 			_str_params.SetNumber(1, this->ci->number_of_trains);
+			_str_params.pluralize_amount = this->ci->number_of_trains;
 			break;
 		case CIW_NUMBER_CARS:
 			_str_params.SetNumber(1, this->ci->cars_per_train);
+			_str_params.pluralize_amount = this->ci->cars_per_train;
```
Alberth289346: Doesn't `Number` imply plural?  Or maybe add `SetCount` for adding plural? (or optional boolean whether plural is wanted)

Point is, most numbers will need a plural line, so you can either add many code lines for that, or hide it in the call.

src/language.cpp
```diff
@@ -35,7 +35,31 @@ TextString::TextString()
...
+
+	uint plural = _language.GetPluralFormIndex(_current_language, amount);
+	if (plural < this->languages[_current_language].size() && this->languages[_current_language].at(plural) != nullptr) {
+		return this->languages[_current_language].at(plural);
+	}
+
+	plural = _language.GetPluralFormIndex(LANG_EN_GB, amount);
+	if (plural < this->languages[LANG_EN_GB].size() && this->languages[LANG_EN_GB].at(plural) != nullptr) {
+		return this->languages[LANG_EN_GB].at(plural);
+	}
```
Alberth289346: Any reason why you don't hide this in `this->languages[LANG_EN_GB].getString(amount)` ?

At some point you will need gender as well so this isn't the last thing.

graphics/rcd/lang/sv_SE.yml
```diff
@@ -0,0 +1,459 @@
...
+		INBOX_TITLE:              "Inkorg"
+		MESSAGE_NEW_RIDE:         "En ny åkattraktion är nu tillgänglig: %1%"
+		MESSAGE_SCENARIO_WON:     "Grattis, du har vunnit scenariot!"
+		MESSAGE_SCENARIO_LOST:    "Du har förlorat scenarios. Bättre lycka nästa gång."
+		MESSAGE_BROKEN_DOWN:      "%1% har gått sönder."
+		MESSAGE_REPAIRED:         "%1% har reparerats."
+		MESSAGE_CRASH_NO_DEAD:    "%1% kraschade!"
+		MESSAGE_CRASH_WITH_DEAD:
+			one:   "%2% gäst dog när %1% kraschade!"
+			other: "%2% gäst(er) dog när %1% kraschade!"
```
Newbytee:
```suggestion
			other: "%2% gäster dog när %1% kraschade!"
```

src/language.cpp
```diff
@@ -35,7 +35,31 @@ TextString::TextString()
 void TextString::Clear()
 {
 	this->name = nullptr;
-	std::fill_n(this->languages, lengthof(this->languages), nullptr);
+	std::fill_n(this->languages, lengthof(this->languages), std::vector<const char*>());
+}
+
+/**
+ * Get the string in the currently selected language.
+ * @param amount The amount for which to return the appropriate plural form.
```
Alberth289346: `amount` expresses an imprecise number in many cases, a quantity that has no exactly known value ("a substantial amount of work"). Here you used exactly known values to find the plural form.

Maybe "value", "count", or "number" or so?

Also only this line talks about plural rather than line 42 and possibly 44 which is about promises made by the function.

"for which" is not proper English. I am guessing it's a literal translation of some German phrase.
You can simplify it like "The value to use for selecting the appropriate plural form."  No need to have "to return" in there. Line 44 is about returned values :)

src/language.cpp
```diff
@@ -233,6 +257,7 @@ void StringParameters::Clear()
 {
 	this->parms.clear();
 	this->set_mode = true;
+	this->pluralize_amount = 1;
```
Alberth289346: This is the number of plural forms?  "count" is much better here then, or "number_plurals" (a bit short though, improvement is possible).

"amount" is very generic notion, beyond "it's a number" it doesn't say much.

Alberth289346: If it is the first plural form, wouldn't be `0` a more logical choice?

src/language.cpp
```diff
@@ -310,14 +339,46 @@ std::string Language::GetText(StringID number)
...
+ * Look up the correct plural form for an amount.
+ * @param lang_index The language to look in.
+ * @param amount The amount to look up.
+ * @return Index of the correct plural form.
+ */
+uint Language::GetPluralFormIndex(int lang_index, int64 amount)
+{
+	if (amount == 1 && this->singular_form_index[lang_index] != -1) return this->singular_form_index[lang_index];
+
+	int result = this->plural_forms[lang_index]->Eval(amount);
```
Alberth289346: I am not sure why you cache(?) singular values.

Is "Eval" that expensive?

It looks like something that shouldn't be needed.

src/language.h
```diff
@@ -99,21 +102,11 @@ class TextString {
...
-		if (this->languages[LANG_EN_GB] != nullptr) return this->languages[LANG_EN_GB];
-		return "<no-text>";
-	}
+	std::string GetString(int64 amount) const;
 
 	/* Memory is not owned. */
-	const char* name;                       ///< Name of the string.
-	const char* languages[LANGUAGE_COUNT];  ///< The string in all languages.
+	const char* name;                                    ///< Name of the string.
+	std::vector<const char*> languages[LANGUAGE_COUNT];  ///< The string in all languages, indexed by plural form.
```
Alberth289346: How messy, different languages have different number of plural forms, and now you can't even push that problem down to the language.

You may want to turn this upside down.
Have proper `Language` that contains `TextPhrase`-es. (Basically "string" but that has too many confusing meanings.)
A TextPhrase may have multiple plural forms.

The current TextString (with a string in all languages) then doesn't exist anymore (although you can construct it again by making it point to all relevant TextPhrase objects), not sure how bad that is.

Lot of work, but I think it will make life more comfortable.

src/language.h
```diff
@@ -155,6 +148,7 @@ struct StringParameters {
 
 	bool set_mode; ///< When not in set-mode, all parameters are cleared on first use of a Set function.
 	std::vector<StringParameterData> parms; ///< Parameters of the string.
+	int64 pluralize_amount;                 ///< Amount to use for plural forms.
```
Alberth289346: "Value to use for selecting the plural form"

src/language.h
```diff
@@ -168,14 +162,21 @@ class Language {
...
+	uint GetPluralFormIndex(int lang_index, int64 amount);
 
 private:
 	/** Registered strings. Entries may be \c nullptr for unregistered or non-existing strings. */
 	const TextString *registered[2048]; // Arbitrary size.
 	uint first_free; ///< 'First' string index that is not allocated yet.
+
+	std::unique_ptr<EvaluateableExpression> plural_forms[LANGUAGE_COUNT];  ///< Compiled expression for every language's plural rule.
+	int nplurals                                        [LANGUAGE_COUNT];  ///< Number of plural forms in each language.
+	int singular_form_index                             [LANGUAGE_COUNT];  ///< The plural form index for amount 1, cached for performance.
```
Alberth289346: Compiled expression fit better in a language class as method. Same with `nplurals`.
Give the compiler a chance to understand these things are used together.

Singular form shouldn't be special enough to warrant caching, I think.

src/rcdgen/string_storage.cpp
```diff
@@ -7,107 +7,292 @@
...
 
 /**
- * Add a collection of strings to the storage.
- * @param strs String collection to add.
- * @param pos Position of the collection.
+ * Parse the provided YAML file and store its translations.
+ * @param filename File to read from.
  */
-void StringsStorage::AddStrings(std::shared_ptr<StringsNode> strs, const Position &pos)
+void StringsStorage::ReadFromYAML(const char *filename)
```
Alberth289346: A function of 250 lines seems excessive.

Can't you break it into smaller pieces? Perhaps use regular expression matching?

src/ride_type.cpp
```diff
@@ -201,17 +201,17 @@ void SetRideRatingStringParam(const uint32 rating)
...
+		str = _language.GetSgText(GUI_RIDE_MANAGER_RATING_LOW);
 	} else if (rating < 500) {
-		str = _language.GetText(GUI_RIDE_MANAGER_RATING_MEDIUM);
+		str = _language.GetSgText(GUI_RIDE_MANAGER_RATING_MEDIUM);
 	} else if (rating < 750) {
-		str = _language.GetText(GUI_RIDE_MANAGER_RATING_HIGH);
+		str = _language.GetSgText(GUI_RIDE_MANAGER_RATING_HIGH);
 	} else if (rating < 1050) {
-		str = _language.GetText(GUI_RIDE_MANAGER_RATING_VERY_HIGH);
+		str = _language.GetSgText(GUI_RIDE_MANAGER_RATING_VERY_HIGH);
```
Alberth289346, line 212: Random entry, but why is singular form so special? Wouldn't 0 parameters do the job?

Alberth289346:
Runtime interpreting expressions for plural form?
Are you afraid a language may change their plural forms?

Introduce an enumeration for the known plural forms, give them a name so you can name them in the yaml file, write CPP code that computes the plural form index for each known form, stick that in the languages that use it.

Build a language factory for each plural form so you can use the plural form for getting a Language object with all the right settings.

Noordfrees:
I think I got all comments now…

- Fixed Swedish typo
- Renamed amount → count
- Hardcoded plural rules and nplurals for each language. Additionally I pulled all language definitions into a single place, instead of having 4 of them.
- Get rid of TextString and use per-language storage.
- Split YAML parsing into a struct with smaller functions.

src/language.cpp
```diff
@@ -254,53 +248,50 @@ void Language::Clear()
...
+		}
+	}
+
+	/* This would be much easier if names and text strings were in the same order. */
+	for (uint i = 0; i < td.string_count; ++i) {
+		assert(names[i] != nullptr && *names[i] != '\0');
+		this->string_names.at(base + i) = names[i];
+		for (uint j = 0; j < td.string_count; ++j) {
+			if (strcmp(names[i], td.strings[j].name) == 0) {
+				for (int l = 0; l < LANGUAGE_COUNT; ++l) this->languages[l].values.at(base + i) = td.strings[j].languages[l];
```
Alberth289346: Wrt the comment, you could build a `std::map<>` from `name` to `j` separately (before line 274), and use that to grab the `j` associated with an `i`.

A little longer, but you gain simplicity and speed here.

src/language.cpp
```diff
@@ -309,24 +300,76 @@ std::string Language::GetText(StringID number)
...
-	if (number < lengthof(this->registered) && this->registered[number] != nullptr) {
-		const std::string &text = this->registered[number]->GetString();
-		if (text.empty()) return "<empty text>";
-		return text;
+	const char *str = languages[_current_language].GetPlural(number, count);
+	if (str != nullptr) return *str != '\0' ? str : "<empty translation>";
+
+	if (_current_language != SOURCE_LANGUAGE) {
+		str = languages[SOURCE_LANGUAGE].GetPlural(number, count);
+		if (str != nullptr) return *str != '\0' ? str : "<empty string>";
```
Alberth289346: Rather than sticking the `if (not the source language)` around here, do it with the previous attempt where you pull the translated string.

If I am playing with the `SOURCE_LANGUAGE`, I won't get the `<empty translation>` in that case.

src/language.cpp
```diff
@@ -309,24 +300,76 @@ std::string Language::GetText(StringID number)
...
  * @param lang_index The language to look in.
  * @return The language name.
  */
-std::string Language::GetLanguageName(int lang_index)
+std::string LanguageManager::GetLanguageName(int lang_index) const
+{
+	return this->languages[lang_index].GetSgText(GUI_LANGUAGE_NAME);
+}
+
+const char *LanguageManager::GetStringName(StringID number) const
```
Alberth289346, line 343: Needs documentation.

src/language.h
```diff
@@ -155,30 +117,48 @@ struct StringParameters {
...
+class LanguageBundle {
+public:
+	LanguageBundle();
+	void Clear();
+	void InitMetaInfo(int index);
+
+	const char *GetSgText(StringID number) const;
+	const char *GetPlural(StringID number, int64 count) const;
+
+	std::vector<PluralizedString> values;  ///< Every known string with all its plural forms, indexed by #StringID.
```
Alberth289346, line 136: Vectors may pre-allocate more entries than you use (to reduce the number reallocations).
For the outer vector this should not be a problem.

For the inner vector, you may want to do something. If you `reserve` for all plural forms (before adding elements), the vector should not grow any more.

Another solution could be an `std::array` as inner container. You get the precise number of entries at minimal space usage then, but the `Language` class becomes templated with the number of plurals (or, I am guessing, the plural form identification that you can use to obtain the number of plurals).  
Not sure that would work.

Another option is to flatten the inner vector into the outer vector by making it have space for all plural forms of each string. To get the right spot you'd need to multiply the string number by the number of plural forms.

Alberth289346: I can see you may want to postpone this until you have more languages.

src/language.h
```diff
@@ -155,30 +117,48 @@ struct StringParameters {
...
+	                                       ///< Strings without a translation are \c nullptr. Memory is not owned.
+	const LanguageDefinition *metadata;    ///< This language's metadata.
+};
+
+/** Class for retrieving language strings. */
+class LanguageManager {
```
Alberth289346, line 142: `Manager` is a bit of a dangerous word, it breathes "everything fits in here" and they may end up as bloated classes that do everything.

Being aware of it is one defense against it. Another trick is to be a bit more precise in what it manages, thus reducing the scope of its name.

src/rcdgen/string_storage.cpp
```diff
@@ -7,107 +7,346 @@
...
+				case ' ':
+					value.push_back(' ');
+					break;
+				case '\"':
+					value.push_back('\"');
+					break;
+				case '\'':
+					value.push_back('\'');
+					break;
+				default:
```
Alberth289346, line 180: `\\` ?

Alberth289346:
Looks much better imho.

I am guessing your yaml decoding works in foreign languages because you skip the bytes of non-ascii codepoints?
(Not a problem, just curious.)

Unless you want to go into templated classes, nothing major it seems.

src/language.h
Noordfrees: Actually the vector will sometimes contain `nplurals` elements and sometimes only 1 element, depending on whether the string in question does have plural forms. So an array and/or templates wouldn't work at all here. However I added a call to `shrink_to_fit` on the inner vectors.

src/language.h
Noordfrees: Good point but I couldn't think of a more descriptive name, can you suggest one?

Noordfrees:
> I am guessing your yaml decoding works in foreign languages because you skip the bytes of non-ascii codepoints?
> (Not a problem, just curious.)

With my system locale set to Low German, which has lots of äöüß, this is easy to test for:
image
Non-ASCII bytes are treated the same way as ordinary letters and copied verbatim.

Noordfrees:
Many thanks for the extensive reviewing :)

PR #274, opened 2022-08-22
Allow sorting the loadsave savegame list

Noordfrees:
image

### DIFF
--- a/src/loadsave_gui.cpp
+++ b/src/loadsave_gui.cpp
@@ -7,6 +7,9 @@
 
 /** @file loadsave_gui.cpp Gui for selecting the file to load from or save to. */
 
+#include <functional>
+#include <list>
+
 #include "stdafx.h"
 #include "window.h"
 #include "gamecontrol.h"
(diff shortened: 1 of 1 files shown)


src/loadsave_gui.cpp
```diff
@@ -15,6 +18,34 @@
 #include "rev.h"
 #include "sprite_store.h"
 
+/** Sorting order for a list of savegames. */
+using SortBy = std::function<bool(const PreloadData&, const PreloadData&)>;
```
LordAro: Generally, std::function should be avoided due to its high overheads and general slowness https://wolchok.org/posts/cxx-trap-2-std-function/

src/loadsave_gui.cpp
```diff
@@ -15,6 +18,34 @@
...
+
+/** Sort by timestamp. */
+static const SortBy SORT_BY_TIMESTAMP = [](const PreloadData &a, const PreloadData &b) {
```
LordAro: Top-level lambdas are a bit... weird. Why not just use actual functions?

src/loadsave_gui.cpp
```diff
@@ -32,12 +63,14 @@ class LoadSaveGui : public GuiWindow {
...
+	void Sort(const SortBy &sort);
 
 private:
 	std::string FinalFilename() const;
 
-	const Type type;                  ///< Type of this window.
-	std::set<PreloadData> all_files;  ///< All files in the working directory.
+	const Type type;                   ///< Type of this window.
+	std::list<PreloadData> all_files;  ///< All files in the working directory.
+	const SortBy *current_sort;        ///< Current sorting order.
```
LordAro: Avoiding std::function does mean you have to use void* to store the function pointer which isn't so nice for the typesystem, but suitable amount of templating (on Sort) should get you pretty close

src/loadsave_gui.cpp
```diff
@@ -32,12 +63,14 @@ class LoadSaveGui : public GuiWindow {
...
 	bool OnKeyEvent(WmKeyCode key_code, WmKeyMod mod, const std::string &symbol) override;
+	void Sort(const SortBy &sort);
 
 private:
 	std::string FinalFilename() const;
 
-	const Type type;                  ///< Type of this window.
-	std::set<PreloadData> all_files;  ///< All files in the working directory.
+	const Type type;                   ///< Type of this window.
+	std::list<PreloadData> all_files;  ///< All files in the working directory.
```
LordAro: std::list ? [Linked list bad](https://stackoverflow.com/a/34171273/995325), just use std::vector

LordAro: std::list ? [Linked list bad](https://stackoverflow.com/a/34171273/995325), just use std::vector

Alberth289346:
Done

PR #275, opened 2022-08-22
Refactor guest block arrays to STL containers

Noordfrees:
Re #212 

I would have liked to do it more elegantly, but this would have broken savegames without some very major intermediate-structure-based redesign.

### DIFF
--- a/src/people.h
+++ b/src/people.h
@@ -11,53 +11,9 @@
 #define PEOPLE_H
 
 #include <list>
+#include <map>
 
 #include "person.h"
-
-static const int GUEST_BLOCK_SIZE = 512; ///< Number of guests in a block.
-
-/** A block of guests. */
-class GuestBlock {
-public:
-	GuestBlock(uint16 base_id);
-
-	/**
-	 * Get a guest from the array.
-	 * @param i Index of the person (should be between \c 0 and #GUEST_BLOCK_SIZE).
-	 * @return The requested person.
-	 */
-	inline Guest *Get(uint i)
-	{
-		assert(i < lengthof(this->guests));
-		return &this->guests[i];
-	}
-
-	/**
-	 * Get a guest from the array.
-	 * @param i Index of the person (should be between \c 0 and #GUEST_BLOCK_SIZE).
-	 * @return The requested person.
-	 */
-	inline const Guest *Get(uint i) const
-	{
-		assert(i < lengthof(this->guests));
-		return &this->guests[i];
-	}
-
-	/**
-	 * Get the index associated with a guest.
-	 * @param g %Guest object to query.
-	 * @return Index of the guest in the block.
-	 */
-	inline uint Index(Guest *g)
-	{
-		uint idx = g - this->guests;
-		assert(idx < lengthof(this->guests));
-		return idx;
-	}
-
-protected:
-	Guest guests[GUEST_BLOCK_SIZE]; ///< Persons in the block.
-};
 
 /**
  * All our guests.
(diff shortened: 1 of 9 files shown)


LordAro:
One of the advantages of the GuestBlock data structure was preallocating extra space (like std::vector does). As far as I'm aware, std::map doesn't do this (being generally implemented as an RB-tree).

Premature optimisation root of all evil, etc etc, but maybe a std::vector would be more efficient for this, though of course would still need to manage all the inactive/active guests

Noordfrees:
The problem with using a vector here is that a lot of other structures (voxels, rides, messages, …) contain pointers to guests, which will all be invalidated when the number of guests grows beyond the preallocated amount of memory. If we want to do that, then we have to set a fixed arbitrary maximum guest count and allocate all of that memory on startup as in master. 
But this means that if your park goes really well, you may eventually hit a completely arbitrary limit beyond which the game decides you have too many guests and refuses to give you more. And if we make that limit really high to mitigate, then we'll block a huge chunk of RAM most of which will most likely never be used.

Although I suppose if it's mostly about memory allocation, then how about a combination of new and old approach, where we reserve small guest blocks (16? 64?) as arrays, and use a map of starting index to guest block for the lookup. Would that be more efficient?

Alberth289346:
/me waves to LordAro*

> Although I suppose if it's mostly about memory allocation, then how about a combination of new and old approach, where we reserve small guest blocks (16? 64?) as arrays, and use a map of starting index to guest block for the lookup. Would that be more efficient?

Having more than 1 block was my first thought. In that way you keep the fixed addresses, and you get dynamic expansion.

Don't know how much data a single guest has, but there is not much point in being conservative I think. My first guess would be at least a 100, possibly even 250 or 500. Not sure how large your average Park is.

Also not sure what the point of a map is. Did you plan to drop some blocks "in the middle" at some point or so? A map introduces the need to get the block entry by hashtable, while you can trivially point at the correct block directly if you (EDIT: Use consecutive guest numbering and) store blocks in a vector.

I wouldn't worry about removing blocks either I think. At best you could drop the highest block if it is empty, but then again, the park needed that block, so  likely it will need it again (the park hasn't suddenly become smaller or so).

Noordfrees:
Let's have the best of two worlds then :) 

I kept allocation batches small at 64 guests. At the current park sizes, it takes quite long to reach that, and then even longer to reach multiples of it, allocating hundreds of guests at a time is a waste of space. 
Also, instead of linear searching for unused slots I now use a vector to cache free indices.

src/people.h
```diff
@@ -133,11 +71,8 @@ class Guests {
 	};
 	Complaint complaints[COMPLAINT_COUNT];  ///< Statistics about all complaint types.
 
-	bool FindNextFreeGuest();
-	bool FindNextFreeGuest() const;
-	bool HasFreeGuests() const;
-	void AddFree(Guest *g);
-	Guest *GetFree();
+	std::vector<std::unique_ptr<Guest[]>> guests;  ///< All guest slots.
+	std::vector<int> free_guest_indices;           ///< Unused indices in %guests.
```
Alberth289346, line 75: I wondered about pre-allocation of `std::vector`, but the `unique_ptr<>` will avoid that problem. Nice!

A bitset is an alternative to store a collection of integers (although for 1000 free guests, you're using 4000 bytes here, not something to worry about :) )  
While reading a bit about `std::bitset` and it's dynamic cousin `std::vector<bool>`, I am missing the "get index of next set bit" function like the Java version has: https://docs.oracle.com/javase/7/docs/api/java/util/BitSet.html#nextSetBit(int)  included since 2002.  
Without that function, the bitset is next to useless for this purpose, so unless you want to write a bitset implementation, you likely already have the better solution :)

PR #277, opened 2022-08-26
Relayout Bottom Toolbar

Noordfrees:
image

It now shows the park rating, and those widgets are clickable to bring up the respective window. Also contains a commit that fixes a crash related to message pluralization – bugs always seem to be discovered directly after the feature that adds them is merged ;)


src/park_management_gui.cpp
```diff
@@ -41,6 +41,7 @@ class ParkManagementGui : public GuiWindow {
  * @ingroup gui_group
  */
 enum ParkManagementWidgets {
+	/* The tabbutton constants must be in sync with #ParkManagementGuiTabs. */
```
Alberth289346: If you add such comments in the doxygen text above, it becomes visible in the documentation as well. Also, stuff like "#ParkManagementGuiTabs" make more sense then.

Not sure if you did that, but add such a note in ParkManagementGuiTabs as well pointing to here.

Of course if you really want to be sure it should be enforced or checked, but that may be a bit too much here.

src/toolbar_gui.cpp
```diff
@@ -358,10 +364,18 @@ Point32 BottomToolbarWindow::OnInitialPosition()
 	return pt;
 }
 
+void BottomToolbarWindow::SetTooltipStringParameters(BaseWidget *tooltip_widget) const
```
Alberth289346, line 367: misses documentation

Alberth289346: Hmm, maybe it's documented in a base class. Not too awake currently :(

src/toolbar_gui.cpp
```diff
@@ -381,14 +395,37 @@ void BottomToolbarWindow::SetWidgetStringParameters(WidgetNumber wid_num) const
...
-	BaseWidget *w = this->GetWidget<BaseWidget>(wid_num);
-	if (pos.x < w->pos.width - w->pos.height) {
-		_inbox.DismissDisplayMessage();
-	} else {
-		_inbox.display_message->OnClick();
+		default:
+			break;
 	}
+
+	return GuiWindow::OnClick(wid_num, pos);
```
Alberth289346: Euhm, why do you return a value when line 396 says "void"?

Alberth289346:
Looks nothing major, so please fix and merge, unless you think I need to check again.

Noordfrees:
Thanks :)

PR #279, opened 2022-08-26
Fix more bugs with uninitialized string parameters

Noordfrees:
Another flavour of the same bug not caught by the previous fix


Alberth289346:
Looks ok

PR #280, opened 2022-08-27
New Ride: Spiral Slide

Noordfrees:
image

No working animation yet, however a spritesheet that does *not* require being sliced into individual voxel stacks.


Alberth289346:
New rides are always useful :)

PR #282, opened 2022-08-28
New Shop: ATM

Noordfrees:
image


Alberth289346:
Nice, now remove the exit, and cash will be pouring in :)
Everything is ok.

Noordfrees:
Not a viable strategy sadly – when a guest decides they want to go home, they stop visiting rides no matter how much money they have left :) 

---

I would design more rides the next days, but I found an annoying problem when designing anything bigger than 2×2: https://freerct.net/forum/post/26
If you happen to have a suggestion how to address this it would be much appreciated :)

Alberth289346:
Oh the famous issue 119 :)

Precedence will fail. Issue 119 is clipping problems in OpenTTD: https://github.com/OpenTTD/OpenTTD/issues/119 where forever new cases appeared while trying to figure out a precedence. Eventually it ended in "wontfix".

It's even possible to construct a non-solvable situation with 3 bars all partly behind another in a cyclic way, so there is no "first" graphics you can draw. I think it's in the issue but it would need checking each image in there. 

Even though your situation is slight different, you have more control about what is being drawn (rather than having to deal with "everything that might happen when vehicles move". However even then I don't think it is the right path. Too much hassle.

I think you should take a look at what Zephyris has done here.
While adding his graphics, I never had to check the position of the tile origin. Each and every image has the exact same XY tile origin position relative to a corner of the image.

If you compare your ship graphics with the carousel graphics by Zephyris, he doesn't do vertical clipping of a tile. it seems to me he is masking all tiles behind away after moving to the correct viewing position for a tile. I never quite understood why he odes that, maybe when tiles in front are below the bottom border or so. It could also be simpler to perform in a render pipeline, no idea.

As far as I know he uses (or used) Blender, so chances are we have the render pipeline as well. He made complete basesets for OpenTTD, so definitely he is not doing that all by hand. As blender is programmable by Python, my guess is he uses that to automatically render each tile.  You basically need to (EDIT: move the viewpoint to the right position to get the desired tile in the center of the result image, and) restrict the model to the area within some cube, which feels feasible to me.

If that fails, another path is to generate output for some render backend like povray. (I only know that one, but there are more, never looked at any of them in more detail so cannot recommend.) Maybe there is more room in such an environment.

Anyway, I think the road is to make a standard setup in blender for positioning and sizing things, and then automate in some way rendering all the tiles by masking some areas away (which expect the standard setup in order to do their work).

PR #283, opened 2022-08-28
Formatting Enhancements

Noordfrees:
Several commits to make formatting functions nicer:
- Localizable date format and guest name.
- Simplified money format function.
- Wrap `s(n)printf` calls in a single `Format()` function, and use `DrawText` instead where applicable.
- Use `error()` instead of `fprintf`+`exit` for error conditions.

### DIFF
--- a/src/language.cpp
+++ b/src/language.cpp
@@ -447,12 +447,10 @@ static void MoneyStrFmt(char *dest, size_t size, double amt)
  * @param temp Temperature in 1/10 degrees Celcius to convert.
  * @return Formatted text.
  */
-static std::string TemperatureStrFormat(int temp)
+static inline std::string TemperatureStrFormat(int temp)
 {
-	temp = ((temp < 0) ? temp - 5 : temp + 5) / 10; // Round to degrees Celcius.
-	static char buffer[64];
-	snprintf(buffer, lengthof(buffer), "%d \u2103", temp);  // " " + degrees Celcius, U+2103
-	return buffer;
+	temp = ((temp < 0) ? temp - 5 : temp + 5) / 10;  // Round to degrees Celsius.
+	return Format("%d \u2103", temp);
 }
 
 /**
--- a/src/language.cpp
+++ b/src/language.cpp
@@ -385,28 +385,39 @@ void LanguageBundle::InitMetaInfo(int index)
  */
 static std::string MoneyStrFmt(int64 money)
 {
-	std::string result;
+	std::string result = _language.GetSgText(GUI_MONEY_CURRENCY_SYMBOL);
 	if (money < 0) {
-		result += "\u2212";  // U+2212 Unicode Minus sign.
+		result += u8"\u2212";  // U+2212 Unicode Minus sign.
 		money *= -1;
 	}
 
 	const std::string tho_sep  = _language.GetSgText(GUI_MONEY_THOUSANDS_SEPARATOR);
 
-	std::string before_thousands = std::to_string(money / 100);
-	int nr_digits = before_thousands.size();
+	const int cents = money % 100;
+	money /= 100;
 
-	result += _language.GetSgText(GUI_MONEY_CURRENCY_SYMBOL);
+	/* Split string into groups of thousands. */
+	int tgroups = 0;
+	int64 temp_money = money;
+	for (; temp_money >= 1000; ++tgroups, temp_money /= 1000);
+	result += std::to_string(temp_money);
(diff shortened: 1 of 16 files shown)


LordAro:
`Format` feels like it might have issues due to returning a(n implicitly constructed) temporary and so forth. Have you checked it?

Noordfrees:
Yes, it's safe. The `char buffer[]` is implicitly converted to `std::string` in the return statement, which makes a deep copy of the buffer's contents. The return value is then implicitly moved, remaining valid. (It also remains valid when the string is copied, which always performs a deep copy.)

`std::string` is very safe and convenient in this regard, not at all like C-style `char*` where memory scope and memory ownership need to be checked all the time.

src/language.cpp
```diff
@@ -369,93 +368,58 @@ void LanguageBundle::InitMetaInfo(int index)
...
+ * @return Formatted money string.
  */
-static void MoneyStrFmt(char *dest, size_t size, double amt)
+static std::string MoneyStrFmt(int64 money)
 {
-	const std::string curr_sym = _language.GetSgText(GUI_MONEY_CURRENCY_SYMBOL);
-	size_t curr_sym_len = curr_sym.size();
+	std::string result;
+	if (money < 0) {
+		result += "\u2212";  // U+2212 Unicode Minus sign.
```
Alberth289346: I'd say this needs an UTF-8 encoding setting somewhere. That is, different compilers or OSes may do different things here.

Afaik `std::string` does not promise UTF-8 in the standard.

src/language.cpp
```diff
@@ -369,93 +368,58 @@ void LanguageBundle::InitMetaInfo(int index)
...
-	uint curpos = 0;
-
-	/* Also has the added bonus of 'removing' the
-	 * automagically included '-' symbol by snprintf */
-	if (amt < 0) {
-		dest[j++] = '-';
-		curpos++;
+	for (int d = 0; d < nr_digits; ++d) {
+		result += before_thousands.at(d);
+		if (d > 0 && d + 1 < nr_digits && (nr_digits - d) % 3 == 1) result += tho_sep;
```
Alberth289346: If you don't want the `before_thousands` temporary, you could divide by 100 (and keeping it around), then recursively divide by 1000 until you reach the end. Then add currency symbol, optional minus sign, and unfold the recursion, appending the intermediate triplets to the string with a thousands separator between them.

Then finally add a decimal dot, and the remainder of the 100 division.

Not sure if you can do this easily, but printf formatter command `%03d` automatically prefixes zeroes upto the specifed length.

That said, this seem ok too.

src/language.cpp
```diff
@@ -369,93 +368,58 @@ void LanguageBundle::InitMetaInfo(int index)
...
-
-	for (uint i = curpos; i < len && j + max_append_size < size; i++) {
-		if (len - i > 3 && (int)(i - comma_start) % 3 == 0) {
-			strncpy(dest + j, tho_sep.c_str(), tho_sep_len + 1);
-			j += tho_sep_len;
-			dest[j++] = buf[i];
+	money %= 100;
+	result += _language.GetSgText(GUI_MONEY_DECIMAL_SEPARATOR);
+	result += '0' + (money / 10);
+	result += '0' + (money % 10);
```
Alberth289346: Perhaps only when there is a non-zero remainder?

It does make alignment a bit more tricky though in columns, although you have that problem anyway if the font has digits with different widths.

src/language.cpp
Noordfrees: C++ does promise to use `char` internally, and that `char` can hold any UTF8 character. What is not promised is that string literals are UTF8-encoded by default, so I prefixed this with `u8""` which guarantees correct UTF8 encoding.

Btw we already use `\u2103` for the temperature string and it has always worked fine. (Well rendering it never worked, but that's a problem with SDL_ttf which might finally go away when we move to OpenGL and Freetype… (edit: and does not affect `\u2212`))

src/language.cpp
Noordfrees: Having numbers constantly switch around between zero, one, or two digits after the comma/period would look inconsistent and weird to me, especially since nobody uses values like "€ 4.1" in practice – it's always either "€ 4" or "€ 4.10". I'd prefer for the number of digits to remain fixed at showing the two-digit cents value.

Alberth289346: Oh sorry, I was intending to say "when money is non-zero", I agree that "4.1" for 4.10 would be weird.

PR #284, opened 2022-08-29
Ride/Scenery Internal Type Names

Noordfrees:
As more rides are added, using type indices for saveloading becomes fragile. Assign every type of ride, entrance, and scenery a unique (unlocalizable) name set by the RCD file for saveloading.


src/sprite_store.cpp
```diff
@@ -1533,7 +1533,9 @@ const char *SpriteManager::Load(const char *filename)
 			if (!shop_type->Load(&rcd_file, sprites, texts)) {
 				return "Shop type failed to load.";
 			}
-			_rides_manager.AddRideType(std::move(shop_type));
+			if (!_rides_manager.AddRideType(std::move(shop_type))) {
+				return "Shop type could not be added.";
```
Alberth289346, line 1537: I can imagine you eventually want the internal name here to limit your search.
Arguably, returning an error message is not the optimal way to report such things :)

PR #286, opened 2022-08-30
New Ride: Top Spin

Noordfrees:
image

(Continued from #282:) I had a look at the carousel blend file and it is already using the same voxel-wise splitting technique I am using for ship swing, spooky manor, and now top spin. Some technical details differ (Zephyris uses cleverly placed planes, I use a boolean intersect modifier) but other than that it's the same. The really big problem is writing up the RCD files so that every individual render is blitted in the correct voxel, for every rotation (those are especially evil), for every frame.
However after a lot of trial and error (because that's what it always is) I found a configuration that mostly works for me (unlike the ship swing's, which results in nicer rcdgen code but is a pain to model as well as to get working) and enhanced the spritesheet generator to automatically emit the rcdgen animation definitions which can then be copy-pasted into the TXT files. The diff looks accordingly frightening – but if such code is acceptable then I suppose I could live with this workflow.

The next two steps that will be needed in this case are:
- A preprocessor for the rcdgen input files that allows splitting this stuff across multiple files. As the files grow, they become harder and harder to edit.
- Generating the RCD file now takes several minutes instead of several seconds. I want to investigate which step it is that takes so long (reading the images or writing them?), and how to speed it up.


utils/GenerateSpritesheet.java
```diff
@@ -96,7 +190,52 @@ public static void main(String[] args) throws Exception {
...
+					"width_x:" + sprites_x + ";" +
+					"width_y:" + sprites_y + ";" +
+					"tile_width:64;" +
+					"(se_{vert(0.." + (sprites_x - 1) + ")}_{hor(0.." + (sprites_y - 1) + ")}):sheet{" +
+						"x_base:" + f + "*" + sprites_x + "*" + sprite_w + ";y_base:0*" + sprites_y + "*" + sprite_h + "+" + sprite_h + "*2;" +
+						"x_step:" + sprite_w + ";y_step:-" + sprite_h + ";" +
+						"file:\"" + out_path + "\";" +
+						"recolour:\"" + mask_path + "\";" +
+						"x_offset:-32;y_offset:32-" + sprite_h + ";width:" + sprite_w + ";height:" + sprite_h + ";" +
+					"}" +
```
Alberth289346: I am not sure why you stick so much onto a single line. Whitespace isn't expensive.

Insert a newline character after every line that you have here, and add normal white space. It makes a world of difference in readability with very little extra effort.

It saves a lot of time if someone ever wants or needs to read this data.

Also a diff will be much more readable. In its current form you absolutely will not understand what you changed between versions. With lines that roughly fit without wrapping your chances are much better.

If that breaks copy/pasting, write it to a file instead of dumping it on the screen? Might be useful too if you can split the source file into pieces.

Alberth289346:
Pasted image seems to have some darker lines running along the base, that may need some checking?

Alberth289346:
Last but not least, I may be away a few days near the end of the week,

Noordfrees:
The dark lines are an artefact of the splitting. It is nearly impossible to perform that with pixel-perfect precision; either the resulting images are a bit too small and there is empty (dark) space between them, or they are too big and some parts of the image are cut out. Another reason why single-images would be better. Since this ride has more than nine thousand individual frames, re-rendering which would take around six or seven hours, and we need *more* rides more importantly than *perfect* rides, I don't want to fix this now; though I'll be try to be more careful with the next ride.

I changed the generator to now save nicely formatted code to a file. When I implement the preprocessing I'll also reformat the Top Spin's animation definitions properly, but for now I'd like to leave those as they are since this long catch-all file is hard to work with reasonably.

Edit: Here's an example for the new autogenerated code. It's 1697 lines just for the working animation. Starting and stopping would be longer.
[rcdcode.txt](https://github.com/FreeRCT/FreeRCT/files/9455437/rcdcode.txt)

Alberth289346:
> Here's an example for the new autogenerated code.

Thanks, I like it.

> The dark lines are an artefact of the splitting. It is nearly impossible to perform that with pixel-perfect precision

So perhaps splitting before rendering is wrong? Would rendering the entire thing and then split into pieces by copying "some" of the pixels onto the image of each tile work?  It should be possible to do that without artefacts by copying either the full pixel or make it fully transparent.
In that way the same pixel in the ride at different tile images is always the same.

> we need more rides 

I agree, so let's leave this ride as is. It can be re-computed once the kinks in the render pipeline are resolved.

For an include mechanism, there is cpp (C pre-processor) with its `#include`. Not ideal probably but readily available.

Another path that _may_ exist is partial "compilation". Rather than doing everything in one rcdgen process, process eg each ride separately, and then either allow a bunch of rcd files to be loaded or merge the results into one big rcd file.
I don't remember what I did there. I did once consider the idea that as a player you could have files from different sources, so someone could make a single ride, but no idea if that works at all.

Alberth289346:
Splitting into tile images after rendering the full ride would need mask images I guess?

PR #287, opened 2022-08-31
RCD Generation Efficiency Enhancements

Noordfrees:
RCDgen input files are now organized in the file types `FML` (FreeRCT Markup Language, passed to the preprocessor) and `FMS` (FreeRCT Markup Snippet, `#include`d in other `FML` or `FMS` files). The `FML` files are processed by the C Preprocessor into `FPP` (FreeRCT PreProcessed file) files, which are then passed to RCDgen.

Most input files have been split into a directory tree of smaller snippets, each of which corresponds to one unit. Classes of similar units are grouped in sub-directories. The includes are structured for readability and ease of modding by having one `main.fms` file in each intermediate directory which just includes all subdirectories. 
The Top Spin has been split into many files, and future rides with similar animations will be organized in the same way.

Additionally, RCDgen now caches images in memory. This reduces RCD generation time from 130 seconds to 20 seconds. This is still much longer than it was before the Top Spin with its large animation was added, but reasonable for now. In the future we may want to check if there are other performance bottlenecks we can improve.

Fun fact: Distributing custom/inofficial RCD files to add single units or sets of arbitrary units is, and has already been, easily possible. Just call the RCDgen pipeline manually on your custom file and dump the output file next to the other RCDs. There are no restrictions about which blocks may or may not be defined in any given RCD file. I haven't tried it, but there is no reason why it shouldn't work – it's essentially what we've been doing whenever we added a new RCD file.


Alberth289346:
Likely it's a good idea to document the new file structure and file extensions in the rcdgen documentation, as here it will become lost very quickly.

I trust final results are equivalent, and glad you managed to trim down execution time of rcdgen.

If you like to document the file structure in a next patch, I am fine with merging this as-is. In that way you can continue in the new setup, and docuementation can be reviewed later.

Noordfrees:
Thanks. I already have the next new ride ready whose branch is stacked on this one, so I'll merge now and write up the documentation afterwards.

PR #288, opened 2022-08-31
New Ride: Chair Swing

Noordfrees:
image


Alberth289346:
This looks really nice!!  This looks much better.

The ride looks a bit cramped, the chairs at the back look like they are outside the ride area, You may want to add more tiles around (this or) the next ride. People should be able to walk around the ride too to enter or leave.

I leave it open so you can decide if you want to change the tiles, I am fine with either decision.

Feel free to merge with or without changing the ride.

Noordfrees:
I had thought I had spaced the seats so they just fit, but when I opened the blend file to make a screenshot as proof I realized the motion actually does move the seats too far to the outside :P 
image image

I'll scale the chairs down a bit and reduce their inner spacing to ensure there's enough empty space around.

Noordfrees:
image

PR #289, opened 2022-09-01
Document FML/FMS/FPP Language Specification

Noordfrees:


developer_documentation/rcdgen_markup.rst
```diff
@@ -0,0 +1,1096 @@
...
+Recommended File System Layout
+==============================
+The following section is a recommendation on how to layout the files. It is permitted to ignore it.
+
+An FML file is a top-level file. It must not be included in any other file.
+It should generate exactly one RCD file by providing exactly one top-level value list which is a `file`_ node.
+Except for the filename extension, the name of the generated file should be identical to the FML file's name.
+
+An FMS file is not preprocessed by itself. It is always included from another FMS or FML file.
+Each FMS file is included only one file.
```
Alberth289346:
```suggestion
Each FMS file is included in only one other file.
```
Alternatively "Each FMS file is included only once." although it may be a bit less clear.

Alberth289346:
Ohh, lots of text even, I only expected the file tree structure rules and filename extension conventions.

Two high level things to consider I think:

- You currently describe c-preprocessing as part of the rcdgen input format. The other option is to consider c-pre-processing as a separate (pre-processing) step, that rcdgen doesn't know about. The latter leads to a more separate description of the pre-processing step and the rcdgen processing step. (EDIT: This is what C/C++ does, the input language description of C never even discusses `#` lines. Only in the cpp section that is explained (and there you wont find anything about the C input language).

  Note you also don't mention the java program as processing step nor do you describe it's input language, while in theory you see it as a 3rd processing step before c-preprocessing, thus generating the cpp files `on the fly` before kicking of cpp.

  I think the decision what to do here is a bit dependent on future extension of rcdgen. If it eventually will read the current file tree (or something close to it) directly (thus eliminating cpp as separate step), then it makes sense to explain the various `#` lines as recdgen input language (we use cpp as a temporary implementation then). If rcdgen will not know about cpp anytime 'soon', then the need to see the cpp lines as rcgen input language is less needed.

- I am not sure you want to go into such detail about the blocks. It basically describes a lot of detail of all blocks which means this text will need changing with every change of a block contents. It may be better to keep a more high level description (eg just a list of blocks that exist, and the general syntax), and then refer to the detailed tables for reader that want to know what is really inside each bock.

Noordfrees:
I changed the text to clarify that FPP is the language accepted by RCDgen, and that FML/FMS are extensions of this standard that can be used optionally to autogenerate FPP files.

I think documenting the attributes accepted by RCDgen is even more important than documenting the RCD file layout. Nobody is going to write an RCD file by hand, and it's unlikely that anyone will write a custom RCD generator or another program that reads RCD files, so this file format is actually closely tied in to FreeRCT internals. On the other hand, FML/FPP is the language that will be used by content designers to create new assets and should have an accordingly detailed documentation. 
Expecting content designers to read the C++ source of `check_data.cpp` to find out why their FPP code doesn't compile is much less developer-friendly than giving them a real, readable documentation file that lists and describes all the attributes as well as any restrictions placed on them. 
And every time a new feature is added we already add a new version of the RCD block with a layout change there. This already involves updates to `nodes.h`, `nodes.cpp`, `check_data.cpp`, `data_format.rst`, *and* the respective RCD loading code – one more file to update is not that much by this comparison.

PR #292, opened 2022-09-06
Move to OpenGL + GLFW + Freetype

### LINKED ISSUE #135: Blit32bppImage Segfaults Intermittently
The function has segfaulted twice on me, but it's inconsistent.

When I was fixing the SDL problem, I was able to discover blit segfaults are due to dereferencing `src_base`. This could mean that `mem` is being reallocated at a different timev than the variables that control `src_base`. I don't really understand why pointer math is being used here, but it _usually_ works flawlessly.

The first time it segfaulted I was moving the FreeRCT window around my desktop, maximizing it, minimizing it, etc. I tried briefly to make it segfault again, and was unsuccessful. If I had to guess, I'd say that `mem` was resized and `src_base` was not made aware of this because of some flaw in how the frame generation works, but the nature of the flaw means the window resize has to hit at exactly the right part of the tick cycle. It's kind of a guess.

The second time it segfaulted was when I built a roller coaster to space and it got too high. Dr. MinGW tells me it was Blit32bppImage; I would have otherwise thought it was not related to blitting. I haven't looked to see how this came about.

Since the second segfault, I'm not so sure my theory about window resizing at the exact right part of the update is actually the problem.

### LINKED ISSUE #223: `heap-buffer-overflow` in `Blit8bppImages`
When dragging a window with a horizontal scrollbar (e.g. the scenery placement window) so that the scrollbar is below the lower edge of the FreeRCT window, the program reproducably crashes with
```
=================================================================
==19207==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7f00fb99aed4 at pc 0x55d6ede62d8b bp 0x7ffc40289810 sp 0x7ffc40289800
READ of size 4 at 0x7f00fb99aed4 thread T0
    #0 0x55d6ede62d8a in Blit8bppImages /home/benedikt/Programmierung/FreeRCT/src/video.cpp:680
    #1 0x55d6ede645b7 in VideoSystem::BlitImages(Point<int> const&, ImageData const*, unsigned short, unsigned short, Recolouring const&, GradientShift) /home/benedikt/Programmierung/FreeRCT/src/video.cpp:823
    #2 0x55d6edc8a0f7 in VideoSystem::BlitImage(Point<int> const&, ImageData const*, Recolouring const&, GradientShift) /home/benedikt/Programmierung/FreeRCT/src/video.h:137
    #3 0x55d6ede85447 in ScrollbarWidget::Draw(GuiWindow const*) /home/benedikt/Programmierung/FreeRCT/src/widget.cpp:865
    #4 0x55d6ede8d758 in IntermediateWidget::Draw(GuiWindow const*) /home/benedikt/Programmierung/FreeRCT/src/widget.cpp:1507
    #5 0x55d6ede8846c in BackgroundWidget::Draw(GuiWindow const*) /home/benedikt/Programmierung/FreeRCT/src/widget.cpp:1168
    #6 0x55d6ede8d758 in IntermediateWidget::Draw(GuiWindow const*) /home/benedikt/Programmierung/FreeRCT/src/widget.cpp:1507
    #7 0x55d6ede9592a in GuiWindow::OnDraw(MouseModeSelector*) /home/benedikt/Programmierung/FreeRCT/src/window.cpp:476
    #8 0x55d6ede9a5ba in WindowManager::UpdateWindows() /home/benedikt/Programmierung/FreeRCT/src/window.cpp:1147
    #9 0x55d6ede9a848 in WindowManager::Tick() /home/benedikt/Programmierung/FreeRCT/src/window.cpp:1168
    #10 0x55d6edca3b2f in OnNewFrame(unsigned int) /home/benedikt/Programmierung/FreeRCT/src/gamecontrol.cpp:76
    #11 0x55d6ede623af in VideoSystem::MainLoop() /home/benedikt/Programmierung/FreeRCT/src/video.cpp:541
    #12 0x55d6edca2cfd in freerct_main(int, char**) /home/benedikt/Programmierung/FreeRCT/src/freerct.cpp:213
    #13 0x55d6ede9c44a in main /home/benedikt/Programmierung/FreeRCT/src/unix/main_unix.cpp:22
    #14 0x7f0113c6b564 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x28564)
    #15 0x55d6edc0e51d in _start (/home/benedikt/Programmierung/FreeRCT/build/bin/freerct+0x17451d)

0x7f00fb99aed4 is located 2772 bytes to the right of 1920000-byte region [0x7f00fb7c5800,0x7f00fb99a400)
allocated by thread T0 here:
    #0 0x7f0114424717 in operator new[](unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:102
    #1 0x55d6ede6035b in VideoSystem::SetResolution(Point<int> const&) /home/benedikt/Programmierung/FreeRCT/src/video.cpp:251
    #2 0x55d6ede61f95 in VideoSystem::HandleEvent() /home/benedikt/Programmierung/FreeRCT/src/video.cpp:516
    #3 0x55d6ede623bb in VideoSystem::MainLoop() /home/benedikt/Programmierung/FreeRCT/src/video.cpp:545
    #4 0x55d6edca2cfd in freerct_main(int, char**) /home/benedikt/Programmierung/FreeRCT/src/freerct.cpp:213
    #5 0x55d6ede9c44a in main /home/benedikt/Programmierung/FreeRCT/src/unix/main_unix.cpp:22
    #6 0x7f0113c6b564 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x28564)

SUMMARY: AddressSanitizer: heap-buffer-overflow /home/benedikt/Programmierung/FreeRCT/src/video.cpp:680 in Blit8bppImages
Shadow bytes around the buggy address:
  0x0fe09f72b580: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0fe09f72b590: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0fe09f72b5a0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0fe09f72b5b0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0fe09f72b5c0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0x0fe09f72b5d0: fa fa fa fa fa fa fa fa fa fa[fa]fa fa fa fa fa
  0x0fe09f72b5e0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0fe09f72b5f0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0fe09f72b600: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0fe09f72b610: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0fe09f72b620: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==19207==ABORTING
make[3]: *** [CMakeFiles/run.dir/build.make:76: CMakeFiles/run] Fehler 1
make[2]: *** [CMakeFiles/Makefile2:163: CMakeFiles/run.dir/all] Fehler 2
make[1]: *** [CMakeFiles/Makefile2:170: CMakeFiles/run.dir/rule] Fehler 2
make: *** [Makefile:205: run] Fehler 2
```
Version: 7aceb948 with ASan enabled
Possibly related to #135

### LINKED ISSUE #291: 32bpp Image Data Layout Is Messed Up
<details>
<summary>TL;DR</summary>
To reproduce, add the following code at the end of `ImageData::Load32bpp`, then compile with ASan enabled and try to start.

```c++
	for (int x = 0; x < this->width; ++x) {
		for (int y = 0; y < this->height; ++y) {
			uint32 colour = this->GetPixel(x, y);
		}
	}
```

I guarantee that one of the first images loaded will cause a crash with a heap-buffer-overflow. It seems to happen at random locations, anywhere inside the image. Adding log data before every `GetPixel` call in the snippet above, the first output after starting is:
```
[100%] Built target freerct
NOCOM 32bpp [ 47x 29]   0x  0
NOCOM 32bpp [ 47x 29]   0x  1
NOCOM 32bpp [ 47x 29]   0x  2
=================================================================
==117653==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x617000000b1f at pc 0x5607b284738a bp 0x7ffdc4c8b6c0 sp 0x7ffdc4c8b6b8
READ of size 1 at 0x617000000b1f thread T0
    #0 0x5607b2847389 in ImageData::GetPixel(unsigned short, unsigned short, Recolouring const*, GradientShift) const /home/benedikt/frct/freerct/src/sprite_data.cpp:201:20
    #1 0x5607b2848182 in ImageData::PostLoad() /home/benedikt/frct/freerct/src/sprite_data.cpp:253:26
    #2 0x5607b2848549 in LoadImage(RcdFileReader*) /home/benedikt/frct/freerct/src/sprite_data.cpp:284:8
    #3 0x5607b28594d3 in SpriteManager::Load(char const*) /home/benedikt/frct/freerct/src/sprite_store.cpp:1363:21
    #4 0x5607b285bdfc in SpriteManager::LoadRcdFiles() /home/benedikt/frct/freerct/src/sprite_store.cpp:1582:10
    #5 0x5607b26e411b in freerct_main(int, char**) /home/benedikt/frct/freerct/src/freerct.cpp:212:18
    #6 0x5607b290f9e1 in main /home/benedikt/frct/freerct/src/unix/main_unix.cpp:22:7
    #7 0x7ff6b8483d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #8 0x7ff6b8483e3f in __libc_start_main csu/../csu/libc-start.c:392:3
    #9 0x5607b25828f4 in _start (/home/benedikt/frct/freerct/build/bin/freerct+0x1568f4) (BuildId: 19917b9e83e8f595ebdebd02563d739fe31a45b4)

0x617000000b1f is located 261 bytes to the right of 666-byte region [0x617000000780,0x617000000a1a)
allocated by thread T0 here:
    #0 0x5607b264061d in operator new[](unsigned long) (/home/benedikt/frct/freerct/build/bin/freerct+0x21461d) (BuildId: 19917b9e83e8f595ebdebd02563d739fe31a45b4)
    #1 0x5607b2845e31 in ImageData::Load8bpp(RcdFileReader*, unsigned long) /home/benedikt/frct/freerct/src/sprite_data.cpp:52:19
    #2 0x5607b2848487 in LoadImage(RcdFileReader*) /home/benedikt/frct/freerct/src/sprite_data.cpp:280:9
    #3 0x5607b28594d3 in SpriteManager::Load(char const*) /home/benedikt/frct/freerct/src/sprite_store.cpp:1363:21
    #4 0x5607b285bdfc in SpriteManager::LoadRcdFiles() /home/benedikt/frct/freerct/src/sprite_store.cpp:1582:10
    #5 0x5607b26e411b in freerct_main(int, char**) /home/benedikt/frct/freerct/src/freerct.cpp:212:18
    #6 0x5607b290f9e1 in main /home/benedikt/frct/freerct/src/unix/main_unix.cpp:22:7
    #7 0x7ff6b8483d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16

SUMMARY: AddressSanitizer: heap-buffer-overflow /home/benedikt/frct/freerct/src/sprite_data.cpp:201:20 in ImageData::GetPixel(unsigned short, unsigned short, Recolouring const*, GradientShift) const
Shadow bytes around the buggy address:
  0x0c2e7fff8110: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2e7fff8120: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2e7fff8130: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2e7fff8140: 00 00 00 02 fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2e7fff8150: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0x0c2e7fff8160: fa fa fa[fa]fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2e7fff8170: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2e7fff8180: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2e7fff8190: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2e7fff81a0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2e7fff81b0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==117653==ABORTING
```
Depending on which image is loaded first, the invalid access may occur in a different line in `GetPixel`. The memory access is always a small number of bytes after the end of the allocated data, sometimes just off by very few bytes, sometimes several lines of ASan shadow bytes.

The crashes only happen in 32bpp images; 8bpp images seem to be fine.

For reference, here's the `GetPixel` implementation:
https://github.com/FreeRCT/FreeRCT/blob/d3af2c4d05c0dc1bcb428e13d223a64352fe5e5c/src/sprite_data.cpp#L160-L243
And the file format documentation:
https://github.com/FreeRCT/FreeRCT/blob/d3af2c4d05c0dc1bcb428e13d223a64352fe5e5c/developer_documentation/data_format.rst#L186-L244

Or maybe the code emitted by RCDgen is at fault?
https://github.com/FreeRCT/FreeRCT/d3af2c4d05c0dc1bcb428e13d223a64352fe5e5c/master/src/rcdgen/image.cpp
</details>

---

**So my question is**,
does anyone understand the format of our 32bpp `ImageData` structure, what is done wrong in `ImageData::GetPixel` for 32bpp sprites, and/or how to fix the out-of-bounds memory access?

**Background** is that I am now seriously tackling the move to OpenGL/OpenGLES, and its rendering mechanism means that, for every sprite loaded from an RCD file in our own image encoding implementation, I need to create a pixel array `uint8[w * h * 4]` once after loading and fill it in with the RGBA values of every pixel in the image, which will afterwards allow rendering the image far more efficiently than the *(about-to-be-removed, hello #223 and #135)* `Blit8|32bppImages` functions ever could. This issue is the last bigger obstacle in the way to completing my graphics rewrite, everything after that will be comparatively easy.

---

Probably related: #135

Noordfrees:
Finally, here's the big, big move to OpenGL. My requirement is OpenGL 3.3, which I *hope* isn't unrealistically high (no idea what a reasonable expectation is, though it *might* work with older versions as well, don't know), and I *think* it's OpenGL ES – at least I trimmed it down until it compiles in WebAssembly which has a very limited feature set.

I decided to go the whole way and not just replace the software rendering but all of SDL. The new backend is GLFW which is a) easier to use, b) offers the same range of features, and c) appears a bit less bug-prone to me (though that might just be because until now I worked with SDL much more than with GLFW).

The font rendering now uses Freetype. I had to include a different default font though because our previous one lacks important characters.

The framerate says it all. Master gets extremely laggy (i.e. unplayable) the moment you run it at anything larger than 800×600. My benchmark for this branch is at 1920×1080, in a debug build with ASan, the desired maximum framerate of 33.3 with no windows open and ~20 FPS with lots of windows.

I'd appreciate if this could get some testing on other systems. Also some thoughts by @PSVita-Dev would be appreciated.

image

Fixes #135 
Fixes #223 
Fixes #291


Alberth289346:
Very nice indeed!

I compiled and ran the program, so that works.

Playing a bit with it gave me 2 problems.
- It's very hard to toggle open/closed, I think it's not about this change, so I'll make a new issue about it.

- Is the tooltip bug new?
  Screenshot from 2022-09-07 07-18-03

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
+	glBindTexture(GL_TEXTURE_2D, 0);
+	FT_Done_Face(face);
+	FT_Done_FreeType(ft);
+
+	/* Check that we have at least a bearing character and a glyph for invalid characters. */
+	std::string sample_text = {BEARING_CHARACTER};
+	const char *c = sample_text.c_str();
+	size_t i = 1;
+	this->GetFontGlyph(&c, i);
+	this->GetFontGlyph(&c, i);
```
Alberth289346: This is weird, calling the same thing twice, perhaps add a comment about it?

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
+		if (length > 0) {
+			++*text;
+			++length;
+		}
+		/* Fall though to default glyph selection. */
+	} else {
+		*text += bytes_read;
+		length -= bytes_read;
+		if (codepoint <= MAX_CODEPOINT && this->characters[codepoint].valid) return this->characters[codepoint];
+		/* Fall though to default glyph selection. */
```
Alberth289346: Insert an empty line before or write `/* else fall though to default glyph selection. */` to make it more obvious that this doesn't happen always.

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
+void TextRenderer::Draw(const std::string &text, float x, float y, float max_width, uint32 colour, float scale)
 {
-	if (this != &cr) {
-		this->absx = cr.absx;
-		this->absy = cr.absy;
-		this->width = cr.width;
-		this->height = cr.height;
-		this->address = cr.address; this->pitch = cr.pitch;
+	if (text.empty()) return;
+	glUseProgram(this->shader);
```
Alberth289346, line 205: Insert empty line before, as you handle a different case now.

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
+	glUniform1f(glGetUniformLocation(this->shader, "text_colour_r"), FGetR(colour));
+	glUniform1f(glGetUniformLocation(this->shader, "text_colour_g"), FGetG(colour));
+	glUniform1f(glGetUniformLocation(this->shader, "text_colour_b"), FGetB(colour));
+	glUniform1f(glGetUniformLocation(this->shader, "text_colour_a"), FGetA(colour));
+
+	glActiveTexture(GL_TEXTURE0);
+	glBindVertexArray(this->vao);
+
+	y += 0.75f * FONT_PADDING_V * this->font_size * scale;
+	x += FONT_PADDING_H * 0.5f;
```
Alberth289346, line 219: Do these magic values need a constant or a comment?

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
 
 /**
- * Initialize the video system, preparing it for use.
- * @param font_name Name of the font file to load.
- * @param font_size Size of the font.
- * @return Error message if initialization failed, else an empty text.
+ * Called by OpenGL when the window size changes.
+ * @param window Window whose size changed.
+ * @param w New width.
+ * @param h New height.
```
Alberth289346: To make it more clear they are new, change their name. Code like `_video.width = w;` may be interpreted as doing nothing if you miss these two lines.

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
-	} else {
-		printf("Could not set window icon (%s)\n", SDL_GetError());
-	}
+	_window_manager.RepositionAllWindows(_video.width, _video.height);
+	NotifyChange(WC_BOTTOM_TOOLBAR, ALL_WINDOWS_OF_TYPE, CHG_RESOLUTION_CHANGED, 0);
+}
 
-	SDL_StartTextInput(); // Enable Unicode character input.
+/**
+ * Called by OpenGL when a key was pressed.
```
Alberth289346: Line 326 contradicts this promise.

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
+ * @param window Window in which the event occurred.
+ * @param key Constant of the pressed key.
+ * @param scancode Scancode of the pressed key.
+ * @param action Whether the mouse button was pressed or released.
+ * @param mods Bitset of modifier keys.
+ */
+void VideoSystem::KeyCallback(GLFWwindow *window, int key, [[maybe_unused]] int scancode, int action, int mods)
+{
+	assert(window == _video.window);
+	if (action == GLFW_RELEASE) return;
```
Alberth289346: Alternatively, you could return if `action != GLFW_PRESS` (I assume). That excludes any other key event that may happen as well.

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
+
+	switch (button) {
+		case GLFW_MOUSE_BUTTON_RIGHT:
+			_window_manager.MouseButtonEvent(MB_RIGHT, action == GLFW_PRESS);
+			break;
+		case GLFW_MOUSE_BUTTON_LEFT:
+			_window_manager.MouseButtonEvent(MB_LEFT, action == GLFW_PRESS);
+			break;
+		case GLFW_MOUSE_BUTTON_MIDDLE:
+			_window_manager.MouseButtonEvent(MB_MIDDLE, action == GLFW_PRESS);
```
Alberth289346, line 500: I have seen people use `GLFW_RELEASE` here, allowing you to either press at the wrong spot and move towards it before release, or have a change of mind after pressing, and move away before release.

Mostly useful for non-mouse systems though, so it's fine to keep this now imho.

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
+	if (this->window == nullptr) {
+		glfwTerminate();
+		error("Failed to open GLFW window\n");
+	}
 
-		case SDL_USEREVENT:
-			return true;
+	glfwMakeContextCurrent(this->window);
+	if (glewInit() != GLEW_OK) error("Failed to initialize GLEW\n");
+	this->UpdateClip();
```
Alberth289346: `error` bails out too right?

Maybe insert an empty line between these to denote it's not one contiguous block?

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
-	while (_video.MainLoopDoCycle());
+	while (this->MainLoopDoCycle());
 }
 
-/**
- * Perform one cycle of the main loop.
- */
-/* static */ void VideoSystem::MainLoopCycle()
+/** Perform one cycle of the main loop. */
+void VideoSystem::MainLoopCycle()
```
Alberth289346: It's still static:
```
src/video.h
97:	static void MainLoopCycle();
```

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
-	/* Opaque, but colour adjusted depending on the old pixel. */
-	return MakeRGBA(r >> 8, g >> 8, b >> 8, OPAQUE);
+/**
+ * Set the current clipping area.
+ * @param rect New clipping area.
+ * @see #PopClip
+ */
+void VideoSystem::PushClip(const Rectangle32 &rect)
+{
+	this->clip.push_back(rect);
```
Alberth289346, line 694: Perhaps check width and height are more than `0` ?

Alberth289346: Reverse coord computing will crash otherwise.

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
 
 /**
- * Blit pixels from the \a spr relative to \a img_base into the area.
- * @param pt Base coordinates of the sprite data.
- * @param spr The sprite to blit.
- * @param numx Number of sprites to draw in horizontal direction.
- * @param numy Number of sprites to draw in vertical direction.
- * @param recolour Sprite recolouring definition.
- * @param shift Gradient shift.
+ * Load an OpenGL shader.
```
Alberth289346: It loads two shaders, maybe this is not about loading shaders but about configuring shaders for OpenGL?

(ie a higher level aim)

src/video.cpp
```diff
@@ -5,861 +5,908 @@
...
+	glShaderSource(vertex_shader_id, 1, &vertex_source_pointer, NULL);
+	glCompileShader(vertex_shader_id);
+
+	/* Check Vertex Shader. */
+	glGetShaderiv(vertex_shader_id, GL_COMPILE_STATUS, &result);
+	glGetShaderiv(vertex_shader_id, GL_INFO_LOG_LENGTH, &info_log_length);
+	if (info_log_length > 0) {
+		std::vector<char> vertex_shader_error_message(info_log_length+1);
+		glGetShaderInfoLog(vertex_shader_id, info_log_length, NULL, &vertex_shader_error_message[0]);
+		error("Compile error in shader %s: %s\n", vp, &vertex_shader_error_message[0]);
```
Alberth289346, line 818: Shouldn't you check `result`?  Ie, is it defined that log is empty on successful compile and non-empty on failed compile for all OpenGL systems that exist?

src/video.cpp
```diff
@@ -911,188 +955,158 @@ void VideoSystem::GetNumberRangeSize(int64 smallest, int64 biggest, int *width,
...
+ */
+void VideoSystem::DoDrawPlainColours(const std::vector<Point<float>> &points, uint32 col) {
+	struct PerVertexData {
+		float gl_x;
+		float gl_y;
+		float r;
+		float g;
+		float b;
+		float alpha;
+	};
```
Alberth289346, line 1035: And now you hope C++ doesn't insert padding in the structure.

src/video.cpp
```diff
@@ -911,188 +955,158 @@ void VideoSystem::GetNumberRangeSize(int64 smallest, int64 biggest, int *width,
...
+	std::vector<PerVertexData> vertices;
```
Alberth289346, line 1038: You could initialize capacity to avoid re-allocation.

src/video.h
```diff
@@ -5,186 +5,277 @@
...
+		GLuint texture_id;   ///< The OpenGL texture used to render this glyph.
+		Point16 size;        ///< Size of this glyph in pixels.
+		Point16 bearing;     ///< Alignment offset from the baseline.
+		GLuint advance;      ///< Horizontal spacing.
+		bool valid = false;  ///< If \c false, all data in this struct is invalid.
+	};
+
+	const FontGlyph &GetFontGlyph(const char **text, size_t &length) const;
+
+	FontGlyph characters[MAX_CODEPOINT + 1];  ///< All character glyphs in the current font indexed by their unicode codepoint.
```
Alberth289346, line 76: This could explode in size for foreign languages, not sure how much of a problem that is now (could improve in the future).

src/video.cpp
Noordfrees: This would have to be implemented in the WindowManager. This is actually not trivial because it breaks movement initiation. Maybe something for a future branch.

src/video.h
Noordfrees: Currently we only take characters from the lower unicode ranges (ASCII, character variants like äöüñç, some symbols like the °C character) and a few from the very high ranges. The space in-between is already being reserved here so we can index all characters by unicode codepoint, so the glyphs there are not loaded but memory is allocated for them. A lot of foreign scripts actually are in this range so the extra memory consumption will be zero.

The reason why I don't *load* these glyphs though is speed. Loading them all takes about five seconds on startup which is excessive. If/when we get more non-latin-script languages we may want to switch to on-demand loading of character ranges.

Noordfrees:
The tooltip bug isn't new, but I added a commit to fix it (this was an easy one).

src/video.cpp
Alberth289346: I don't see how you addressed this.

Non-empty log does not imply failed compile, see https://stackoverflow.com/questions/56923918/gl-info-log-length-returns-1#56924187

Alberth289346:
Deciding a shader gives errors purely on log size seems to allow for false positives, as far as I can see. Likely that should be addressed.

EDIT: Other stuff seems fine to me. Not sure how much testing should be done before merge?

Noordfrees:
Okay I had misunderstood that, this should address it. 

I think testing doesn't need to be very in-depth, OpenGL usually either works fine or has really glaring bugs such as inverted colours or failure to start. Smaller bugs will be found by and by, and they tend to appear only after merging anyway ;)

PR #297, opened 2022-09-09
Draw float-away money amounts on construction

Noordfrees:
https://user-images.githubusercontent.com/48293446/189337969-4e718f4c-8129-498c-9c9a-313bd2d4f7f8.mp4

### DIFF
--- a/src/viewport.cpp
+++ b/src/viewport.cpp
@@ -861,6 +861,17 @@ void PixelFinder::CollectVoxel(const Voxel *voxel, const XYZPoint16 &voxel_pos, 
 }
 
 /**
+ * Constructor with default speed and fade settings.
+ * @param text Text to show.
+ * @param pos Pixel position in the world.
+ * @param col RGB colour of the text.
+ */
+FloatawayText::FloatawayText(std::string text, XYZPoint32 pos, uint32 col)
+: FloatawayText(text, pos, XYZPoint32(3, -6, 0), col | 0xff, 3)
+{
+}
+
+/**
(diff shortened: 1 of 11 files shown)


src/viewport.cpp
```diff
@@ -917,6 +928,19 @@ void Viewport::ToggleUndergroundMode()
...
+ * @param money Amount of money paid.
+ * @param voxel Voxel coordinate where to draw the text.
+ */
+void Viewport::AddFloatawayMoneyAmount(const Money &money, const XYZPoint16 &voxel)
+{
+	_str_params.SetMoney(1, -money);
+	this->floataway_texts.emplace_back(DrawText(STR_ARG1),
+			XYZPoint32(voxel.x * 256 + 128, voxel.y * 256 + 128, voxel.z * 256 + 128),
+			money < 0 ? 0x00ff0000 : money > 0 ? 0xff000000 : 0x00000000);
+}
```
Alberth289346, line 943: Skip if you neither pay nor get any money?

Alberth289346:
Money should flow up instead of sideways?

LGTM, please merge after considering my remarks

Noordfrees:
I like sideways better: OpenTTD lets costs float straight up, which has the side-effect that the text is obscured by the cursor if the mouse is being moved upwards during the action. A slightly diagonal sideways motion is less likely to cause that problem. Also, it's how RCT1 did it:

https://user-images.githubusercontent.com/48293446/189376286-d0568ad5-a9be-4cc0-a80b-3a946792051d.mp4

PR #298, opened 2022-09-10
Ride Construction & Removal Enhancements

Noordfrees:
- All rides now cost money to build.
- Pricing of elevated platforms now depends on height.
- Removing a ride returns some money. Rides lose value slowly.
- More useful construction GUI for all rides and especially coasters.

image image image


Alberth289346:
Hmm, we need a free build mode :)

Looks good to me, not sure if this can be merged without the other patch, so feel free to decide the order.

Noordfrees:
> Hmm, we need a free build mode :)

We already have it: Menu → Editor
It removes all artificial restrictions such as checking for land ownership and enough money. However in the not-so-distant future I intend to split this off into a separate game mode started from the main menu via a "Scenario Editor" button, instead of being able to toggle it in-game.

A mode with unlimited money like some scenarios in RTC1+LL, and a true sandbox mode like in RCT3, is also on my wishlist but with less priority.

PR #299, opened 2022-09-10
Make certain buttons repeating and text input widgets clickable

Noordfrees:


src/video.cpp
```diff
@@ -260,13 +260,14 @@ void TextRenderer::Draw(const std::string &text, float x, float y, float max_wid
 /**
  * Estimate the bounding rectangle of a string.
  * @param text Text to estimate.
+ * @param padding Include text padding in the result.
  * @param scale Scaling factor for the text size.
  */
-PointF TextRenderer::EstimateBounds(const std::string &text, float scale) const
+PointF TextRenderer::EstimateBounds(const std::string &text, bool padding, float scale) const
```
Alberth289346:
```suggestion
PointF TextRenderer::EstimateBounds(const std::string &text, bool add_padding, float scale) const
```
or `include_padding` but that's a bit long.

ie rather than saying "space" say "add-space", making it more clear what the flag is doing from, its name.

src/video.cpp
```diff
@@ -489,19 +494,16 @@ void VideoSystem::ScrollCallback([[maybe_unused]] GLFWwindow *window, [[maybe_un
 void VideoSystem::MouseClickCallback([[maybe_unused]] GLFWwindow *window, int button, int action, [[maybe_unused]] int mods) {
 	assert(window == _video.window);
 
+	MouseButtons mouse;
```
Alberth289346: Name is far from optimal, a `mouse` is not just a button. It also has a position and sometimes 4 legs. That said, the alternative `button` is already used in the header so that is not an option either. (It's also a bit too general I think, but probably not enough to fix now.)

Likely a name that is close enough is and not used already is `mouse_button`. How does that sound to you?

src/video.cpp
```diff
@@ -945,11 +947,12 @@ void VideoSystem::GetNumberRangeSize(int64 smallest, int64 biggest, int *width,
  * @param text Text to calculate.
  * @param width [out] Resulting width.
  * @param height [out] Resulting height.
+ * @param padding Include text padding in the result.
```
Alberth289346: More padding

src/viewport.cpp
```diff
@@ -1303,8 +1303,9 @@ void Viewport::OnMouseMoveEvent(const Point16 &pos)
 	}
 }
 
-WmMouseEvent Viewport::OnMouseButtonEvent(MouseButtons state)
+WmMouseEvent Viewport::OnMouseButtonEvent(MouseButtons state, WmMouseEventMode mode)
 {
+	if (mode != WMEM_PRESS) return WMME_NONE;  // Repeated buttons should be ignored for all actions below.
```
Alberth289346: Not sure "repeated buttons" is accurate. That is only the case if you don't press a second mouse button while holding the first mouse button.

Perhaps "Avoid processing multiple events for the same click?" (ie you get more events rather than more buttons).

There is also a connection to drag-detection, but that may be more confusing than clarifying.

src/widget.cpp
```diff
@@ -694,10 +706,44 @@ void TextInputWidget::SetText(const std::string &text)
...
+bool TextInputWidget::OnClick([[maybe_unused]] const Point32 &base, const Point16 &pos)
+{
+	this->has_focus = true;
+
+	/* Find clicked cursor pos. \todo We might be able to use a more efficient logarithmic search. */
+	int w;
+	int text_len = this->buffer.size();
+	int best_x = INT_MAX;
+	for (int p = text_len; p >= 0; --p) {
+		_video.GetTextSize(this->buffer.substr(0, p), &w, nullptr, false);
```
Alberth289346: Depending on what magic the font renderer does, this might give wrong intermediate positions due to kerning.

Likely this doesn't affect us currently, as we paint each character separately? If so, another path for your 'todo' is to add character widths incrementally then. For current processors, that may even be the quickest way as they favor sequential access rather than jumping around (and then do sequential access for summing the widths) when you do bisection or so.

src/widget.cpp
```diff
@@ -694,10 +706,44 @@ void TextInputWidget::SetText(const std::string &text)
...
+	return true;
+}
+
 bool TextInputWidget::OnKeyEvent(WmKeyCode key_code, WmKeyMod mod, const std::string &symbol)
 {
-	if (!this->visible) return false;
+	if (!this->visible || !this->has_focus) return false;
 	switch (key_code) {
+		case WMKC_CANCEL:
+			this->has_focus = false;
```
Alberth289346, line 748: I wonder if you should to that for ENTER as well. For windows that disappear after finishing it makes no difference, but if the window is not closed, any further key press will change the last confirmed text-box now?

src/widget.h
```diff
@@ -104,6 +105,7 @@ class BaseWidget {
 	uint16 fill_y;             ///< Vertical fill step.
 	uint16 resize_x;           ///< Horizontal resize step.
 	uint16 resize_y;           ///< Vertical resize step.
+	bool repeating_events;     ///< Receive repeating events.
```
Alberth289346: Name doesn't mean much to me. Explanation doesn't clarify it either.

What is a "repeating event"? Getting the same event multiple times? To me, a press event and a release event are not the same, if that is what you mean. They are significantly different, eg in drag detection.

src/widget.h
```diff
@@ -242,6 +246,7 @@ class TextInputWidget : public LeafWidget {
 	size_t cursor_pos;                ///< Position of the cursor in the text.
 	int value_width;                  ///< Width of the image or the string.
 	int value_height;                 ///< Height of the image or the string.
+	bool has_focus;                   ///< The widget currently owns the keyboard focus.
```
Alberth289346: You're writing explanations that don't explain anything, basically `bool X; // (do/get) X;`

The explanation should explain X without using X, as people reading the comments wouldn't know what X means (or they wouldn't read the comments).

src/window.cpp
```diff
@@ -775,15 +774,16 @@ void WindowManager::AddToStack(Window *w)
 /**
  * Remove a window from the list.
  * @param w Window to remove.
+ * @param about_to_be_deleted The window will be deleted soon after.
  */
-void WindowManager::RemoveFromStack(Window *w)
+void WindowManager::RemoveFromStack(Window *w, bool about_to_be_deleted)
```
Alberth289346: Not sure about the flag here. The name seems to imply the window is deleted somewhat, doesn't it?

Wrong use could lead to crashes due to the `current_window` pointing to a deleted window.

Perhaps there should be several functions here, one for each use?
Then you can move this flag inside the window manager code without exposing it publically. Alternatively, this function could crash if you try to remove a current window (making it the responsibility of the caller to handle it)

src/window_constants.h
```diff
@@ -87,6 +87,16 @@ enum WmMouseEvent {
 	WMME_MOVE_WINDOW,  ///< Initiate a window move.
 };
 
+/**
+ * Mouse event modes of the window manager.
+ * @ingroup window_group
+ */
+enum WmMouseEventMode {
+	WMEM_PRESS,    ///< Mouse button was pressed.
+	WMEM_REPEAT,   ///< Mouse button was held down.
```
Alberth289346: Minor suggestion:

```suggestion
	WMEM_REPEAT,   ///< Mouse button is held down.
```
?

(It's an internal event based on current state, rather than an external event received from the event queue.)

src/viewport.cpp
Alberth289346: I ran into the explaining comment at the end of the review, so my comments may not be entirely optimal :p  

Still not convinced about "repeated button" though, perhaps "auto-repeat"?  That expresses more that it happens all by itself (for me at least). Keyboard also has auto-repeat iirc.

src/viewport.cpp
```diff
@@ -1305,7 +1305,7 @@ void Viewport::OnMouseMoveEvent(const Point16 &pos)
 
 WmMouseEvent Viewport::OnMouseButtonEvent(MouseButtons state, WmMouseEventMode mode)
 {
-	if (mode != WMEM_PRESS) return WMME_NONE;  // Repeated buttons should be ignored for all actions below.
+	if (mode != WMEM_PRESS) return WMME_NONE;  // Avoid processing multiple events for the same click for all actions below.
```
Alberth289346: This was my idea due to not knowing what a repeated button was, you may want to rephrase it?

src/window.cpp
```diff
@@ -771,19 +771,27 @@ void WindowManager::AddToStack(Window *w)
 	stack->lower = w;
 }
 
+/**
+ * Called by a window prior to its deletion.
+ * @param w Window being deleted.
+ */
+void WindowManager::PreRemove(Window *w)
+{
+	if (this->current_window) this->current_window = nullptr;
```
Alberth289346: Shouldn't it compare with `w`?
That was what the previous version did I think.

I am also pondering on the name `PreDelete` is perhaps better, or `Unmanage` :p  
I can live with `PreRemove` too though, any of those names seems fine.

Alberth289346:
The comparison may need fixing, and perhaps a change of the function name if you like a different one better.

Otherwise LGTM, feel free to merge after considering the above.

PR #301, opened 2022-09-14
Clear more clang-tidy hits and protect against regressions

Noordfrees:


src/dates.cpp
```diff
@@ -52,9 +52,8 @@ Date::Date() : day(1), month(1), year(1), frac(0)
  * Copy constructor.
  * @param d Existing date.
  */
-Date::Date(const Date &d) : day(d.day), month(d.month), year(d.year), frac(d.frac)
-{
-}
+Date::Date(const Date &d)
+= default;
```
LordAro, line 56: I'd just leave these in the header, tbh

PR #303, opened 2022-09-25
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/ for the definition of this variable.

This patch was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

### DIFF
--- a/src/rcdgen/check_data.cpp
+++ b/src/rcdgen/check_data.cpp
@@ -11,7 +11,9 @@
 #include <map>
 #include <vector>
 #include <memory>
+#include <cstdlib>
 #include <ctime>
+#include <sstream>
 #include "ast.h"
 #include "nodes.h"
 #include "string_storage.h"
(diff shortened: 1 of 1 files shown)


src/rcdgen/check_data.cpp
```diff
@@ -2810,6 +2812,15 @@ static std::shared_ptr<INFOBlock> ConvertINFONode(std::shared_ptr<NodeGroup> ng)
 
 	char buffer[32];
 	time_t cur_time = time(nullptr);
+	char *source_date_epoch = std::getenv("SOURCE_DATE_EPOCH");
+	if (source_date_epoch) {
```
Noordfrees:
```suggestion
	const char *source_date_epoch = getenv("SOURCE_DATE_EPOCH");
	if (source_date_epoch != nullptr) {
```

src/rcdgen/check_data.cpp
```diff
@@ -2810,6 +2812,15 @@ static std::shared_ptr<INFOBlock> ConvertINFONode(std::shared_ptr<NodeGroup> ng)
...
+		std::istringstream iss(source_date_epoch);
+		iss >> cur_time;
+		if (iss.fail() || !iss.eof()) {
+			fprintf(stderr, "Error: Cannot parse SOURCE_DATE_EPOCH as integer\n");
+			exit(27);
```
Noordfrees: Not sure this should be a fatal error, perhaps just print the warning and fall back to current time?

Edit: Also better include the malformed `source_date_epoch` in the log message

bmwiedemann:
Thanks. I integrated your suggestions.

PR #304, opened 2022-09-26
Zoom

Noordfrees:
The game can now be zoomed. Any sprites not available at a given zoom scale are downscaled from the smallest bigger size, if available, otherwise upscaled from the largest smaller size. All animations and other sprite types support having separate graphics for each scale. 

I did not actually add any sprite definitions for scales other than 64 yet; that's something for a follow-up-branch even for those units where we already have high-resolution renders.

Scaling is done in software. I tried out using OpenGL scaling but the pro-con table is pretty clear:

<table>
<tr>
	<th>
	<th>Software
	<th>Hardware
<tr>
	<th>Pros
	<td><li>Pretty fast after a few frames due to caching
		<li>Images get pixelated but not blurred
		<li>Images align nicely at the edges
	<td><li>Easy to code
<tr>
	<th>Cons
	<td><li>Requires lots of memory
	<td><li>Images get blurred, which strains the eyes
		<li>Terrible quality around image edges (black lines everywhere)
		<li>Very slow
		<li>Very high CPU load
</table>

To address the RAM issue, I'm using a transient cache that drops cached scales that haven't been accessed in a while when the cache size exceeds a threshold.
For comparison, there's also a branch https://github.com/Noordfrees/FreeRCT/tree/zoom-with-hardware-scaling.

---

grafik
grafik


developer_documentation/data_format.rst
```diff
@@ -612,31 +612,38 @@ Version history
...
+=================  =======  =======  =================================================================
+
+A view consists of a sprite block reference for each tile and zoom scale of the object.
+The index order from minor to major is x - y - z.
 
 Version history
 ...............
 
 - 1 (20210131) Initial version.
+- 2 (20220915) Add zoom.
```
Alberth289346, line 646: There are 2 ways to add multiple sizes of a building I think (at least in theory).

One way is what you did, extend the graphics of a building to include multiple zooms of the same building.
The other way is to have several FSET blocks, each with a different zoom.

Your solution assumes that the same author makes all zoom variants. In the 'other way' case, a different author can make a new version of the building independently of the first author. (Again, in theory at least, I haven't checked how feasible that scenario is currently.)

Obviously you can do the latter with your new blocks as well, so nothing is lost here (modulo the feasibility check whether 2 such building definitions would be connected as variations on the same ride while loading).

Alberth289346: With "other author" I mean you get 2 rcd files, one from each author.

Imho if you want to give artists much room to make their own graphics, they shouldn't have to modify data files of other authors. So quite not trivial to achieve.

That all said, your solution will do fine.

developer_documentation/rcdgen_markup.rst
```diff
@@ -592,25 +590,36 @@ CARS
...
+
+Sprites
+.......
+
+If the scales are specified with a single ``tile_width`` parameter, the sprites are named:
+
+- ``car_pPrRyY`` - The sprites for all orientations of the car, for values of *P* (pitch), *R* (roll), and *Y* (yaw) from 0 to 15 each.
+- ``guest_sheet`` - The animation's guest overlay sheet_.
+
+Otherwise, they are named ``car_pPrRyYwW`` and ``guest_sheet_W`` for each tile width ``W``.
```
Alberth289346, line 614: At some point this will need an example I think. As we currently don't have graphics artists, I don't think it's needed now. though.

src/orientation.h
```diff
@@ -95,4 +95,44 @@ static inline bool IsBackEdge(ViewOrientation orient, TileEdge edge)
...
+};
+
+/** Available zoom scales, sorted from smallest to biggest. */
+constexpr ZoomScale _zoom_scales[] = {
+	{ 16,  4},
+	{ 32,  8},
+	{ 64, 16},
+	{128, 32},
+	{256, 64},
+};
```
Alberth289346, line 122: Not sure it's useful to hard-code these.

The factor 4 between the first and second number is from the isometric view, so you never get rid of that. Then the first value is a simple power of 2, although with your zoom idea maybe it's possible to have any width as long as it's a multiple of 4.

But for now power of 2 is enough and probably simplest to implement. It's just that at larger sizes doubling the size is a big step.

src/path.h
```diff
@@ -111,9 +111,11 @@ enum PathType {
 
 /** Path status. */
 enum PathStatus {
-	PAS_UNUSED,      ///< %Path is not loaded.
-	PAS_NORMAL_PATH, ///< %Path to walk on.
-	PAS_QUEUE_PATH,  ///< %Path to queue on.
+	PAS_NORMAL_PATH,  ///< Path to walk on.
```
Alberth289346: I think the `%` is about not linking the name to the `Path` class, you'd have to check the doxygen output to check.

Alberth289346: Shuffling path values doesn't break save games?

src/sprite_data.cpp
```diff
@@ -333,9 +408,76 @@ const uint8 *ImageData::GetRecoloured(GradientShift shift, const Recolouring &re
...
+
+/**
+ * Scale this image to a different size.
+ * @param factor Factor by which to scale.
+ * @return The scaled instance.
+ */
+const ImageData *ImageData::Scale(float factor) const
+{
+	constexpr float EPSILON = 0.01;  ///< Threshold for float comparisons.
+	if (fabs(factor - 1.0f) < EPSILON) return this;
```
Alberth289346: Why don't you use the actual needed sprite pixel size versus the supplied sprite pixel size(s)?

These real-valued factors tend to introduce glitches due to rounding.

src/sprite_data.h
```diff
@@ -13,6 +13,7 @@
 #include <map>
 #include <memory>
 #include "palette.h"
+#include "time.h"
```
Alberth289346: time? what is that doing here?

src/sprite_store.h
```diff
@@ -97,24 +98,13 @@ class Fence : public RcdBlock {
...
-	ImageData *sprites[PATH_COUNT]; ///< %Path sprites, may contain \c nullptr sprites.
-};
-
 /**
  * %Path decoration sprites.
  * @ingroup sprites_group
  */
 class PathDecoration {
 public:
+	static constexpr int LITTER_VOMIT_COUNT = 4;  ///< Number of different litter or vomit sprites.
```
Alberth289346, line 107: I'd expect use of this in the litter/vomit image data below.

src/time.h
```diff
@@ -0,0 +1,37 @@
+/*
+ * This file is part of FreeRCT.
+ * FreeRCT is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 2.
+ * FreeRCT is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with FreeRCT. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file time.h Functions for measuring time. */
```
Alberth289346: Ah right. you may want to rename it to a filename that is not used in the standard library,

Also, I always tried to push as many `.h` files down into the `.cpp` files to avoid including too much for each file. Not sure if that is feasible here as well.

Alberth289346:
To me, the patch is too large, it seems to change too many things at the same time. For you it no doubt makes sense, but for me I constantly have to guess what change belongs to what.

----
What happened I think is that you had "zoom" in mind, then started hacking and changing and more changing and adapting and fixing and when it worked pushed the result to here.

While that works for small changes, for big changes you inevitably always start at the wrong end. While implementing you find it needs more ground work, and other things the get in the way and need a change.
If you read the commit messages you see this happening.

What I learned in hacking OpenTTD is the above (you always start at the wrong end), but also how to deal with it.
My solution to this is to first get a working solution, ie just hack away. Once you have a working solution, look at it again, and see if you can split the patch into independent parts. Second, see if you can modify the order of changes to get a more layered approach. eg you can introduce FSET without zoom and even without support for multiple sizes in the code.

The patch submission path gets longer, but more gradual. Also, the history of changes becomes simpler to trace in the future.
Now, if you find something is broken in this patch after bisecting 5 years from now, then understanding why that particular thing was changed here is very difficult. On the other hand, if each commit has a single focused goal, such reasoning becomes much simpler.

In other words, the idea is to stack as many small patches as you can, building from the ground towards the feature, and then finally add the feature itself once you added all the ground work.

EDIT: Thus basically do the same idea again as a second attempt. As you know what you'll encounter there are less surprises, and with the thinking about patch/change order, patches get more focussed.  
In my experience, I usually wasn't quite happy with the second attempt and I did the same thing again a 3rd time.  
I know it sounds like a lot of work, but these second/third attempts is mostly about copying the right changes into the next patch. Finding out what needs to be changed is what takes a lot of time. Typing the changes is much quicker.

Alberth289346:
I'll leave it to you what to do. I did see some stuff, but nothing major, and obviously it works, so I am fine with merging.

On the other hand, this might be a good time to experiment with how to manage large changes without exploding the patch size.

src/orientation.h
Noordfrees: I changed it to add more scales (heights 24 and 48) and automatic width calculation.

Noordfrees:
Okay, I'll try to rework the commits in a more logical order. This may take another couple of days though.

I do most of my development at @Widelands where we always squash-merge PRs; the master branch always contains 1 opaque commit per PR which contains all the changes as one big batch. In this workflow, "intermediate commits", commit ordering and similar are not something one needs to think about. 
My approach here was to make 1 type of sprite scalable as proof-of-concept, then implement a zoom backend that utilizes it, and then convert all other types of sprites one by one to use it.

Noordfrees:
I've been trying to untangle the commits for some days now and haven't found a commit tree that looks really good. There are a few preparation changes that can easily be pulled out as independent commits before the other changes, but other than that everything else belongs together. It's either a linear progress of smaller, mostly self-contained changesets with intermediary states that don't look that great (like now), or a detailed one-feature-at-a-time progression that would look nicer but can't compile (even worse for bisecting, IMHO), or a single huge catch-all commit "Implement zooming/scaling for all and everything". Big features with fundamental changes on all layers are always evil to put under version control properly… 

So I'm going to merge this as-is after all. Thanks for the review

Alberth289346:
Thanks for trying, pity it didn't work out.

PR #305, opened 2022-11-04
Make use of std::filesystem in fileio operations.

rsf92:
This has the advantage of making it more platform agnostic + reducing code complexity. 
 
Could be related to #212


rsf92:
I pushed changes to fix the clang errors. The macOS errors seem weird, like there's no support of C++17 until macOS 10.15 from 2 years later?

Noordfrees:
Thanks for the patch :) 
Tested and working on Linux.

Regarding the Mac CI, you could edit `.github/workflows/ci-build.yaml` and bump `MACOSX_DEPLOYMENT_TARGET` to 10.15. I guess this should probably fix it.

A couple of codestyle-related nits:
- After keywords like `for, if, while, switch` a space before the opening parentheses is required.
- Similar, add a whitespace before `catch`.
- `if` statements need to be written either on a single line or use `{braces}`.
- Make variables and especially references `const` whenever applicable.
- For `std::string`, prefer `const &` over pass-by-value.
- The `*&` operators attach to the variable name, not the type name.

rsf92:
Okay, i did those changes, plus used `std::filesystem::copy_file` to CopyBinaryFile, which I had, for some reason, forgotten to change.

src/fileio.cpp
```diff
@@ -16,17 +16,18 @@
...
-	#include "windows/fileio_windows.h"
 	#include <direct.h> // contains chdir in windows
 #endif
 #include <sys/types.h>
-#include <sys/stat.h>
-
+#ifdef LINUX
+const char *DIR_SEP = "/";  ///< Directory separator character.
+#elif WINDOWS
+const char *DIR_SEP = "\\";  ///< Directory separator character.
```
Newbytee: Doesn't Windows "tolerate" using a forward slash as directory separator?

rsf92: No idea, it was done like that in the previous implementation. I lack a windows machine so I can't actualliy check that...

Newbytee: I found an even better option: `std::filesystem::path::preferred_separator`.

Newbytee: This article has good information about the subject in general: https://en.cppreference.com/w/cpp/filesystem/path

rsf92: Sounds good, I'll play with it later on and push the changes when I'm done.

src/fileio.cpp
```diff
@@ -16,16 +16,15 @@
...
 	#include <errno.h>
 	#include <unistd.h>
 #elif WINDOWS
-	#include "windows/fileio_windows.h"
 	#include <direct.h> // contains chdir in windows
 #endif
 #include <sys/types.h>
-#include <sys/stat.h>
+
+const char *DIR_SEP = "/";  ///< Directory separator character.
```
Newbytee: How come you're not using `std::filesystem::path::preferred_separator`?

rsf92: You're right, done :)

src/fileio.cpp
```diff
@@ -16,16 +16,15 @@
...
 	#include <errno.h>
 	#include <unistd.h>
 #elif WINDOWS
-	#include "windows/fileio_windows.h"
 	#include <direct.h> // contains chdir in windows
 #endif
 #include <sys/types.h>
-#include <sys/stat.h>
+
+const char DIR_SEP = std::filesystem::path::preferred_separator;  ///< Directory separator character.
```
Newbytee:
```suggestion
constexpr auto DIR_SEP = std::filesystem::path::preferred_separator;
```
I think this comment is superfluous now that we're using a well-documented built-in property of the standard library. Also, using `constexpr` should be fine here. Additionally, I'm not sure about putting `char` here. `std::filesystem::path::preferred_separator` has the type `wchar_t` on Windows. Maybe `auto` makes sense?

src/fileio.cpp
```diff
@@ -51,85 +50,99 @@ const char* LoadingError::what() const noexcept
...
-/**
- * @fn bool DirectoryReader::EntryIsFile()
- * Test whether the last returned entry from #NextEntry is a file.
- * @return Whether the entry is a file.
- */
+	if (!PathIsFile(path)) return;
+	try {
+	    std::filesystem::remove(path);
+	} catch(std::filesystem::filesystem_error const& ex) {
+	        error("%s", ex.what());
```
Newbytee: Inconsistent whitespace?

src/fileio.cpp
```diff
@@ -51,85 +50,99 @@ const char* LoadingError::what() const noexcept
...
 /**
- * Construct a directory reader object (specific for the operating system).
- * @return A directory reader.
- * @ingroup fileio_group
+ * Test whether the given path points to a directory.
+ * @param path Path to investigate.
+ * @return Whether the given path points to a directory.
  */
-DirectoryReader *MakeDirectoryReader()
+bool PathIsDirectory(const std::string &path)
```
Newbytee, line 114: You're putting the ampersand next to the type in some places, and next to the name in other places. I'm not sure how it's done in the code at the moment, but I think it would be good to be consistent about it.

src/fileio.cpp
```diff
@@ -51,85 +50,99 @@ const char* LoadingError::what() const noexcept
...
+	        return false;
+	    }
 }
 
 /**
  * RCD file reader constructor, loading data from a file.
  * @param fname Name of the file to load.
  */
-RcdFileReader::RcdFileReader(const char *fname)
+RcdFileReader::RcdFileReader(std::string fname)
```
Newbytee:
```suggestion
RcdFileReader::RcdFileReader(const std::string& fname)
```
I think this should be fine here.

src/fileio.cpp
```diff
@@ -336,57 +349,35 @@ bool RcdFileReader::GetBlob(void *address, size_t length)
...
-
-#ifdef _WIN32
-	if (CreateDirectory(path.c_str(), NULL)) return;
-	error("Failed creating directory '%s'\n", path.c_str());
-#else
-	if (mkdir(path.c_str(), 0x1FF) == 0) return;
-	error("Failed creating directory '%s' (%s)\n", path.c_str(), strerror(errno));
-#endif
+    bool check = std::filesystem::create_directories(path);
+    if( !check ) {
```
Newbytee:
```suggestion
    if (!check) {
```

src/fileio.cpp
```diff
@@ -336,57 +349,35 @@ bool RcdFileReader::GetBlob(void *address, size_t length)
...
-		int byte = getc(in_file);
-		if (byte == EOF) break;
-		putc(byte, out_file);
+	if (dest.empty() || PathIsFile(dest)) {
+		error("Path is empty or not a file: %s\n", dest.c_str());
 	}
 
-	fclose(in_file);
-	fclose(out_file);
+    std::filesystem::copy_file(src,dest);
```
Newbytee: Inconsistent whitespace with previous lines (there are other occurrences of this as well).

src/fileio.h
```diff
@@ -22,41 +24,15 @@ struct LoadingError : public std::exception {
...
-
-	const char *NextFile();
-	virtual const char *MakePath(const char *directory, const char *fname) = 0;
-
-	virtual bool EntryIsFile() = 0;
-	virtual bool EntryIsDirectory() = 0;
-};
-
-extern const char *DIR_SEP;  ///< Directory separator character.
+extern const char DIR_SEP;  ///< Directory separator character.
```
Newbytee: This won't work if we change the definition to use `auto`. How about just moving the definition from the source file to the header?

rsf92: For some reason the compiler decided that std::filesystem is undeclared there, lol

src/fileio.cpp
rsf92: This should be fixed now

src/fileio.cpp
rsf92: This should be fixed now

src/fileio.h
rsf92: Okay, I figured this out, for some reason rcdgen used C++ 11 still, I updated the CXX_STANDARD for that

rsf92:
Is there any pending review for this PR?

Newbytee:
> Is there any pending review for this PR?

There is a merge conflict. It would be nice if you could resolve that.

rsf92:
> > Is there any pending review for this PR?
> 
> There is a merge conflict. It would be nice if you could resolve that.

Sure, i'll get into it ASAP. I meant apart from it, since it's been open for a few months and maybe I missed some feedback or something.

Newbytee:
I'll take another look. Benedikt Straub said he's been busy with other things  but that he will get back to FreeRCT soon hopefully.

src/rcdfile.cpp
```diff
@@ -73,37 +73,27 @@ void RcdFileCollection::ScanDirectories()
...
+	for (const std::string &rcd_path : _rcd_paths) this->ScanDirectory(rcd_path, 3);
 }
 
 /**
  * Recursively scan a directory, looking for RCD files to add.
  * @param dir Directory to scan.
  * @param recursion_depth Remaining layers of recursion depth.
  */
-void RcdFileCollection::ScanDirectory(const char *dir, int recursion_depth)
+void RcdFileCollection::ScanDirectory(const std::string& dir, int recursion_depth)
```
Newbytee: Ampersand inconsistency here still.

src/rcdfile.cpp
```diff
@@ -145,7 +135,7 @@ static std::string GetString(RcdFileReader &rcd_file, uint max_bytes, uint32 *re
  * @param fname Filename of the file to scan.
  * @return Error message, or \c nullptr if no error found.
  */
-const char *RcdFileCollection::ScanFileForMetaInfo(const char *fname)
+const char *RcdFileCollection::ScanFileForMetaInfo(std::string& fname)
```
Newbytee: Here as well.

src/rcdfile.h
```diff
@@ -30,13 +30,13 @@ class RcdFileCollection {
 	RcdFileCollection();
 
 	void ScanDirectories();
-	void ScanDirectory(const char *dir, int recursion_depth);
+	void ScanDirectory(const std::string& dir, int recursion_depth);
```
Newbytee: Also here.

src/rcdfile.h
```diff
@@ -30,13 +30,13 @@ class RcdFileCollection {
...
 	void AddFile(const RcdFileInfo &rcd);
 
 	std::map<std::string, RcdFileInfo> rcdfiles; ///< Found unique RCD files, mapping of uri to the Rcd file information.
 
 private:
-	const char *ScanFileForMetaInfo(const char *fname);
+	const char *ScanFileForMetaInfo(std::string& fname);
```
Newbytee: And here.

src/loadsave_gui.cpp
```diff
@@ -191,17 +191,12 @@ static const WidgetPart _loadsave_gui_parts[] = {
...
 {
 	/* Get all .fct files in the directory. */
-	std::unique_ptr<DirectoryReader> dr(MakeDirectoryReader());
-	dr->OpenPath((freerct_userdata_prefix() + DIR_SEP + SAVEGAME_DIRECTORY).c_str());
-	for (;;) {
-		const char *str = dr->NextEntry();
-		if (str == nullptr) break;
-		std::string name(str);
+
+	for (const std::string& name : GetAllEntries((freerct_userdata_prefix() + DIR_SEP + SAVEGAME_DIRECTORY))) {
```
Newbytee: This line too.

src/fileio.h
```diff
@@ -22,41 +24,15 @@ struct LoadingError : public std::exception {
...
+constexpr auto DIR_SEP = std::filesystem::path::preferred_separator;  ///< Directory separator character.
 
 /**
  * Class for reading an RCD file.
  * @ingroup fileio_group
  */
 class RcdFileReader {
 public:
-	RcdFileReader(const char *fname);
+	RcdFileReader(const std::string fname);
```
Newbytee:
```suggestion
	RcdFileReader(const std::string &fname);
```
I think this should be fine. Let me know if I'm missing something.

src/fileio.cpp
```diff
@@ -51,85 +48,99 @@ const char* LoadingError::what() const noexcept
...
+ * @param path Path to delete.
  */
+void RemoveFile(const std::string &path)
+{
+
+	if (!PathIsFile(path)) return;
+	try {
+		std::filesystem::remove(path);
+	} catch(std::filesystem::filesystem_error const &ex) {
+		error("%s", ex.what());
```
Newbytee: I'm not sure about catching the error and just printing it here. I think either you should let the caller handle the error (more complicated) or crash the program (easier). This pattern is repeated in some other places.

Newbytee (changes requested):
Overall this looks good, just some concerns I would like to see addressed.

src/fileio.cpp
rsf92: Hmm maybe in void functions we can let the program crash and in  non-void function we return some error value (like false in boolean functions)

Newbytee: One alternative could be to let the exception propagate instead of catching it. Another option (pun unintended) could be to use `std::optional`: https://en.cppreference.com/w/cpp/utility/optional

I think it would be preferable to be consistent about this rather than handling it differently depending on the function's return type.

rsf92: Propagating seems a better solution to me

rsf92:
I don't know why the windows build is failing and only the windows failing. Dropping the const specifier does not seem like the right solution to me, but I have no idea what else to do....

Newbytee:
> I don't know why the windows build is failing and only the windows failing. Dropping the const specifier does not seem like the right solution to me, but I have no idea what else to do....

Could it be that `preferred_separator` has a different type on Windows, and this causes issues?

src/fileio.h
```diff
@@ -22,41 +24,15 @@ struct LoadingError : public std::exception {
...
-
-	const char *NextFile();
-	virtual const char *MakePath(const char *directory, const char *fname) = 0;
-
-	virtual bool EntryIsFile() = 0;
-	virtual bool EntryIsDirectory() = 0;
-};
-
-extern const char *DIR_SEP;  ///< Directory separator character.
+constexpr auto DIR_SEP = std::filesystem::path::preferred_separator;  ///< Directory separator character.
```
Noordfrees: According to the C++ spec, the "character type used by the native encoding of the filesystem: char on POSIX, wchar_t on Windows "
So on Windows this is a different type than on Linux and Mac. You'll need to specify the type as `char` and use some conversion magic.

rsf92: It seems that the DIR_SEP call could be dropped here, since it's already included in the function definition on fileio.cpp

`
const std::string &TrackDesignDirectory()
{
	static std::string dir;
	if (dir.empty()) {
		dir = freerct_userdata_prefix();
		dir += DIR_SEP;
		dir += TRACK_DESIGN_DIRECTORY;
		dir += DIR_SEP;
	}
	return dir;
}
`

Maybe that could be tried and if not working, revert the change and try to do some conversion magic :)

rsf92: I've tested dropping that and things seem to work just fine. I'll see if it compiles and then fix the merge issues

rsf92: It seems that didn't work... which kind of conversion magic would you suggest? It would also be possible to use plain '/' since it's supported by both Windows and Unix

Noordfrees: I suppose a simple cast should do it, with a sanity check to make sure the conversion is valid:

```c++
static_assert(std::filesystem::path::preferred_separator > 0 && std::filesystem::path::preferred_separator <= 127, "Filesystem path separator is not a valid ASCII character");
constexpr const char *DIR_SEP = {static_cast<char>(std::filesystem::path::preferred_separator)};  ///< Directory separator character.
```

Or alternatively we could keep a platform-specific define:

```c++
#ifdef _WIN32
constexpr const char *DIR_SEP = "\\";
#else 
constexpr const char *DIR_SEP = "/";
#endif
```

Newbytee: > Or alternatively we could keep a platform-specific define

This seems like a worse option in my opinion given that it negates the portability gained by using `std::filesystem`.

rsf92: There's a problem here. It seems that wchar_t and char are not directly convertable unless you use specific functions https://stackoverflow.com/questions/36948253/take-wchar-t-and-put-into-char.

I've set it to '\' since it's supposed to work on both windows and unix machines and await futher feedback

src/rcdfile.h
```diff
@@ -30,14 +30,14 @@ class RcdFileCollection {
...
-	void ScanDirectory(const char *dir, int recursion_depth);
+	void ScanDirectory(const std::string &dir, int recursion_depth);
 	void AddFile(const RcdFileInfo &rcd);
 
 	std::map<std::string, RcdFileInfo> rcdfiles; ///< Found unique RCD files, mapping of uri to the Rcd file information.
 	std::vector<std::string>           ftkfiles; ///< Found unique FTK file paths.
 
 private:
-	const char *ScanFileForMetaInfo(const char *fname);
+	const char *ScanFileForMetaInfo(std::string &fname);
```
Noordfrees: Should be
```c++
	const char *ScanFileForMetaInfo(const std::string &fname);
```
(same in the cpp file)

src/rcdfile.cpp
```diff
@@ -74,40 +74,30 @@ void RcdFileCollection::ScanDirectories()
...
-void RcdFileCollection::ScanDirectory(const char *dir, int recursion_depth)
+void RcdFileCollection::ScanDirectory(const std::string &dir, int recursion_depth)
 {
-	std::unique_ptr<DirectoryReader> reader(MakeDirectoryReader());
 
-	reader->OpenPath(dir);
-	for (;;) {
-		const char *fname = reader->NextEntry();
-		if (fname == nullptr) break;
+	for (std::string  filename : GetAllEntries(dir)) {
```
Noordfrees:
```suggestion
	for (const std::string &filename : GetAllEntries(dir)) {
```

src/gamecontrol.cpp
```diff
@@ -115,8 +115,9 @@ void Autosave()
 	/* Roll old autosaves. */
 	for (int i = _max_autosaves - 1; i > 0; --i) {
 		std::string old_file = AutosaveFilename(i);
-		if (PathIsFile(old_file.c_str())) {
-			CopyBinaryFile(old_file.c_str(), AutosaveFilename(i + 1).c_str());
+		if (PathIsFile(old_file)) {
+		    std::string new_file = AutosaveFilename(i + 1);
+			CopyBinaryFile(old_file, new_file);
```
Noordfrees: Inconsistent indentation

src/freerct.cpp
```diff
@@ -125,13 +125,9 @@ static void MigrateOldFiles()
...
 			homedir + DIR_SEP + ".local" + DIR_SEP + "share" + DIR_SEP + "freerct",
 			homedir + DIR_SEP + ".freerct"}) {
-		std::unique_ptr<DirectoryReader> dr(MakeDirectoryReader());
-		dr->OpenPath(old_directory.c_str());
-		for (;;) {
-			const char *filename = dr->NextEntry();
-			if (filename == nullptr) break;
 
-			std::string name(filename);
+		for (std::string name : GetAllEntries(old_directory)) {
```
Noordfrees:
```suggestion
		for (const std::string &name : GetAllEntries(old_directory)) {
```

src/fileio.h
```diff
@@ -22,41 +24,15 @@ struct LoadingError : public std::exception {
...
-
-	const char *NextFile();
-	virtual const char *MakePath(const char *directory, const char *fname) = 0;
-
-	virtual bool EntryIsFile() = 0;
-	virtual bool EntryIsDirectory() = 0;
-};
-
-extern const char *DIR_SEP;  ///< Directory separator character.
+constexpr  auto DIR_SEP = '/';  /// Directory separator character.
```
Noordfrees: Okay, let's go with this then. But please with explicit type:
```suggestion
constexpr char DIR_SEP = '/';  ///< Directory separator character.
```

src/fileio.cpp
```diff
@@ -51,85 +48,86 @@ const char* LoadingError::what() const noexcept
...
- * Set up the directory reader object for reading a directory.
- * @param path Path to the directory.
- * @note Failure is not reported here, but #NextEntry will not return anything useful.
+ * Scan a given directory and return all entries.
+ * @param Directory to scan.
  */
+std::vector<std::string> GetAllEntries(const std::string &path)
+{
+	std::vector<std::string> entries;
+	if(!PathIsDirectory(path)) return entries;
```
Noordfrees: Several whitespace problems in this file: A whitespace is mandatory after the keywords `if` and `for`.

src/fileio.cpp
```diff
@@ -51,85 +48,86 @@ const char* LoadingError::what() const noexcept
 }
 
 /**
- * @fn void DirectoryReader::OpenPath(const char *path)
- * Set up the directory reader object for reading a directory.
- * @param path Path to the directory.
- * @note Failure is not reported here, but #NextEntry will not return anything useful.
+ * Scan a given directory and return all entries.
+ * @param Directory to scan.
```
Noordfrees:
```suggestion
 * @param path Directory to scan.
```

same below

src/fileio.cpp
```diff
@@ -336,57 +334,34 @@ bool RcdFileReader::GetBlob(void *address, size_t length)
...
-		int byte = getc(in_file);
-		if (byte == EOF) break;
-		putc(byte, out_file);
+	if (dest.empty() || !PathIsFile(dest)) {
+		error("Path is empty or not a file: %s\n", dest.c_str());
 	}
 
-	fclose(in_file);
-	fclose(out_file);
+	std::filesystem::copy_file(src,dest,std::filesystem::copy_options::overwrite_existing);
```
Noordfrees: A whitespace is mandatory after every comma

Noordfrees:
Okay, a few more stylistic and performance-related nits (and a clang-tidy hit), but then this should be good to go :)

src/freerct.cpp
rsf92: Dropped const here because the value is modified afterwards. Everything else should be fixed now, including tidy problem :)

src/freerct.cpp
```diff
@@ -125,13 +125,9 @@ static void MigrateOldFiles()
...
 			homedir + DIR_SEP + ".local" + DIR_SEP + "share" + DIR_SEP + "freerct",
 			homedir + DIR_SEP + ".freerct"}) {
-		std::unique_ptr<DirectoryReader> dr(MakeDirectoryReader());
-		dr->OpenPath(old_directory.c_str());
-		for (;;) {
-			const char *filename = dr->NextEntry();
-			if (filename == nullptr) break;
 
-			std::string name(filename);
+		for (std::string &name : GetAllEntries(old_directory)) {
```
Noordfrees: `const`

src/fileio.h
```diff
@@ -22,41 +24,15 @@ struct LoadingError : public std::exception {
...
-
-	const char *NextFile();
-	virtual const char *MakePath(const char *directory, const char *fname) = 0;
-
-	virtual bool EntryIsFile() = 0;
-	virtual bool EntryIsDirectory() = 0;
-};
-
-extern const char *DIR_SEP;  ///< Directory separator character.
+constexpr  char DIR_SEP = '/';  /// Directory separator character.
```
Noordfrees: `///` → `///<`

Noordfrees (commented):
Two more :)

src/freerct.cpp
rsf92: Done

src/fileio.h
rsf92: Done

Noordfrees:
Thanks for this change, and for your patience :)

rsf92:
A pleasure :)

PR #306, opened 2023-03-14
Save and Load Coaster Track Designs

Noordfrees:
Hi, I'm back :)
And here's what I've been working on, with long interruptions, since last autumn:

grafik
grafik

This feature has some room for improvements, such as bundling track designs in an RCD file and the need to rename coasters, and I also want to redesign the selection UI, but it's a fully functional baseplate.


developer_documentation/track_designs.rst
```diff
@@ -0,0 +1,70 @@
+:Author: The FreeRCT Team
+:Version: 2022-10-05
+
+.. contents::
+   :depth: 2
+
+#####################################
+FreeRCT Markup Language Specification
```
Alberth289346: Markup Language Specification??

Not FreeRCT TracK file format or so?

Noordfrees: Oops, copy-paste error, fixed

Alberth289346:
Great addition, and welcome back :)

PR #313, opened 2023-03-23
Speed up RCD file generation by indexing blocks by length

Noordfrees:
Reduces generation of the coaster RCD file from 65s to 29s (54% faster).


Alberth289346:
Ha, great!

Issue #269, opened 2022-06-09
Cannot run prebuilt windows binary(x64)


YJSoft:
Env : Windows 11(22000.675)

I tried

1. 0.1 Installer from `freerct.net`
2. Installer exe from github
3. Portable zip from github

None of above works. just nothing. No error message, or window.

Noordfrees:
Can you try whether manually setting the install dir fixes it? To do so open a command line prompt and run:
```
"C:\Path\To\freerct.exe" --installdir "C:\Path\To\Installation\Directory"
```
The installation directory is the one that should have two folders called `data` and `rcd` in it (e.g. if you extracted the latest ZIP it would be `...\freerct-0.2~git1-c09dc7f-windows-win64\share\freerct`).

RetroSoftwareRepository:
The installer also doesn't come with the required 64-bit DLLs.

LoveBodhi:
I download latest daily build (Portable Zip), still NO required DLL's, even 32-bit!!

PR #316, opened 2023-09-29
Fix Windows linking and install dir detection

Noordfrees:
- Fixes the current failure in the MSVC CI (harfbuzz seems to be not required at all)
- Fix autodetection of the installation directory on Windows (re #269)

The resulting x64 Debug binary was tested in Wine 8.0.1 on Linux where it runs successfully.


Alberth289346 (approved):
Can't say much about it, should be fine I guess

PR #317, opened 2023-10-08
Missions and Scenarios

Noordfrees:
This implements a mission and scenario system as blueprinted in https://freerct.net/forum/topic/8.

The scenario editor is now separate from regular gameplay, with a lot of new UI controls to change scenario settings. It is not as powerful as I would like from an editing point of view, but pretty good for getting started.

Scenarios are grouped into missions. Of each mission only a limited number of scenarios are unlocked, with more becoming available when you first solve a scenario.

I created one default mission with two scenarios. More to come eventually.

grafik
grafik


developer_documentation/rcdgen_markup.rst
```diff
@@ -873,9 +877,22 @@ Attributes:
...
 
+MISN
+~~~~
+Attributes:
+
+- ``internal_name`` - Internal name of the mission.
+- ``texts`` - strings_ node containing the mission's texts.
+- ``max_unlock`` - Maximum number of unlocked unsolved scenarios in the mission.
+- ``scenarios`` - Total number of scenarios in the mission.
+- ``file_N`` - Path to the FCT file for the N-th scenario.
```
Alberth289346, line 903: Should we have requirements wrt to "path"?

- is it an absolute path or a relative path? If the latter, relative to what?
- Kind of directory separators? (unless we have path strings already, but I don't think they exist yet in the rcd files.)

developer_documentation/rcdgen_markup.rst
```diff
@@ -873,9 +877,22 @@ Attributes:
...
+- ``internal_name_N`` - Internal name of the N-th scenario.
+- ``texts_N`` - strings_ node containing the N-th scenario's texts.
+
```
Alberth289346, line 906: Should it have author and/or version of the mission data?

I can imagine an author makes one version and then fixes or improves things, and it seems useful to be able to understand what you get. Similarly, people may copy the mission and adjust it to their liking.

I am also wondering about licenses, can an author define a license of their work?

graphics/rcd/lang/de_DE.yml
```diff
@@ -41,22 +41,27 @@ strings:
...
+		OBJECTIVETEXT_PARK_VALUE:  "Der Park-Verkehrswert muss mindestens %1% betragen."
 
 		NUMBERED_INSTANCE_NAME: "%1% #%2%"
 		RESOLUTION:             "%1%×%2%"
 		GUEST_NAME:             "Besucher %1%"
 
+		NR_DAYS:
+			one:                  "%1% Tag"
+			other:                "%1% Tage"
+		APPLY:                    "Anwenden"
```
Alberth289346, line 53: Middle 2 are somewhat off wrt indenting.

src/gamecontrol.cpp
```diff
@@ -234,54 +270,69 @@ void GameControl::QuitGame()
...
-			item->vox_pos.x = 8 + i;
-			item->vox_pos.y = 1;
-			item->vox_pos.z = (i == 1 ? 12 : 8);
-			_scenery.AddItem(item);
-		}
-		for (int i = 0; i < 6; i++) {
-			BuildFlatPath(XYZPoint16(9, i, 8), PAT_CONCRETE, PAS_NORMAL_PATH, false, false);
+		/* When not loading a scenario, create some default objects. \todo Clean this mess up. */
+
+		_scenario.SetDefaultScenario();
```
Alberth289346: Don't know how you implemented this, but perhaps have a "default scenario" in the RCD files?
Then you don't have this case, and changes are not here but in the data of that scenario.

src/gamecontrol.cpp
```diff
@@ -234,54 +270,69 @@ void GameControl::QuitGame()
...
-void GameControl::StartLevel()
+/**
+ * Initialize common game settings and view.
+ * @oaram editor Launch the game as a scenario editor session.
+ */
+void GameControl::StartLevel(bool editor)
 {
-	_game_mode_mgr.SetGameMode(GM_PLAY);
-	this->speed = GSP_1;
+	_game_mode_mgr.SetGameMode(editor ? GM_EDITOR : GM_PLAY);
```
Alberth289346: I wonder if we get other editors in the future, eg a map editor?  a pre-built ride editor?

`editor` as name is perhaps a bit too generic?

One way out is to make it an enum I guess.

src/mission_gui.cpp
```diff
@@ -0,0 +1,216 @@
...
+				Widget(WT_VERT_SCROLLBAR, SSW_SCROLLBAR, COL_RANGE_BLUE),
+		EndContainer(),
+	});
+
+	this->SetupWidgetTree(scenario_select_gui_parts.data(), scenario_select_gui_parts.size());
+	this->SetScrolledWidget(SSW_MAIN_PANEL, SSW_SCROLLBAR);
+
+	this->SelectTab(this->tab_bar.front());
+}
+
```
Alberth289346, line 101: #undef SCENARIO_ROW_BUTTON   to avoid accidents.

Alberth289346 (approved):
Looking good :)

developer_documentation/rcdgen_markup.rst
Noordfrees: Since it is all wrapped in RCD files this is the same as for all custom content. Versioning can be handled via the `INFO` block (but we only have rudimentary support for that currently). 

Custom licensing is currently not possible, but if desired should also go in the `INFO`. Sometime in the far future we might want to define some contributor and publishing policies about such things. 
Over at @Widelands we decided not to let people do that, and the add-ons server simply has a confirmation checkbox that all add-ons must be published under the GPLv2.

graphics/rcd/lang/de_DE.yml
Noordfrees: Artefact of tabs for indent. When we finally move to Weblate (soon-ish) it will likely break all indenting again anyway and we will no longer maintain these files manually.

PR #321, opened 2025-04-17
Move to GL(ES) 1.x and some QoL

technicaljicama:
Hey there!
This is my first time contributing to FreeRCT. I really like the project so far. 
## I have made a few changes:
- First, I added a little preview in the building gui to see what you are going to place.
- Next is the move to GL(ES) 1.0. This might seem a bit weird, but in my opinion GLES1 usually runs better and allows more compatibility. The WASM build does not like some of it for now, maybe GL3 could be enabled at wish. (GL3 seems overkill for a RCT game).
- I added optional GPerfTools to CMakeLists.txt.
- I fixed a bug where the speed of the coaster cart would still be saved, even after the station, so extreme coasters caused the cart to still have acceleration when leaving the station.
- find_package(glfw) seemed to not work, even though GLFW was installed.
## Some things I'd like to add in the future:
- An "easy" python script to create new (simple) rides as a question/answer system (And an additional blender script for rendering rides).
- Fix right mouse button drag when building a ride.
- Add zoom-in/out buttons in the toolbar directly and rotation buttons
- Scaleable toolbar
- A more memory efficient coaster.
- Ports to multiple platforms.
- More realistic park rating system

I'd love to hear some feedback and let me know your thoughts!

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -40,7 +40,7 @@ IF(WEBASSEMBLY)
 	)
 
 	add_flags(CMAKE_EXE_LINKER_FLAGS "${WASM_FLAGS} \
-			-s FORCE_FILESYSTEM=1 -s ALLOW_MEMORY_GROWTH=1 -s EXIT_RUNTIME=1 -s STACK_SIZE=131072 -s USE_GLFW=3 -s FULL_ES3 -s ASSERTIONS=1 -s WASM=1 -s ASYNCIFY \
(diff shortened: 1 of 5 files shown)


Alberth289346:
Hello!

Nice that you took the time to improve the program.

However, unfortunately you sent your patch to the wrong repository. This repository is dormant. Development is done at https://codeberg.org/freerct/freerct

Not your fault though, it seems. It's not prominently shown here.
I once added a banner for that to the readme, but apparently that has disappeared again :(

To move forward, I think the best thing you can do is visit the above project, and send your patch there.

I glanced through the code, and it seems to do several different things at the same time. It may be a good idea to split different topics into different patches (and different PRs). In that way each topic can be discussed on its own.

@Noordfrees Ideas on how to avoid this from happening again?

